PR #3, opened 2021-11-29
UL - Twitter UI Clone

usamalodhi:
There is a markdown file with titles to the comment headings used in the css styles file for better navigation


index.html
```diff
@@ -0,0 +1,240 @@
...
+          <div class="news-item news-item-4__text">
+            <p>
+              Lorem ipsum dolor sit amet, consectetur adipisicing elit. Deleniti
+              eius doloremque totam nihil commodi accusamus.
+            </p>
+          </div>
+          <div class="image-container__news-1">
+            <img
+              class="boolean-logo"
+              src="../boolean-uk-html-boolean-image-grid/assets/images/boolean.jpeg"
```
Auenc, line 61: There is something a little bit wrong with how you're stating the `src` for the images here. It'll work on your local machine, however, when it comes to viewing this on a different platform (codesandbox for instance) you rimages won't load. to fix this, you would have to change your image `src` to something like `./assets/images/boolean.jpeg`.

index.html
```diff
@@ -0,0 +1,240 @@
...
+            <div class="box-container box-container--right">
+              <i class="far fa-star"></i>
+            </div>
+          </div>
+          <div class="main__top-b">
+            <div class="main__top-b-pic">
+              <img
+                class="avatar"
+                src="../boolean-uk-html-twitter/images/avatar-one.jpg"
+                alt=""
```
Auenc, line 112: It's important to have an `alt` field for every image we use, that way people using devices such as screen-readers can understand what our images are. I believe this is for a profile, so in this case it would be acceptable to have the user's twitter handle/full name if available.

index.html
```diff
@@ -0,0 +1,240 @@
...
+              class="big-pic"
+              src="../boolean-uk-html-twitter/images/big-pic.jpg"
+              alt="big-pic"
+            />
+          </div>
+        </div>
+      </div>
+      <div class="lh-sb">
+        <nav class="nav-grid">
+            <div id="logo" ><a href="https://www.google.com/"><svg class="icon twitter" xmlns="http://www.w3.org/2000/svg" data-name="Layer 1" viewBox="0 0 24 24"><path d="M22,5.8a8.49,8.49,0,0,1-2.36.64,4.13,4.13,0,0,0,1.81-2.27,8.21,8.21,0,0,1-2.61,1,4.1,4.1,0,0,0-7,3.74A11.64,11.64,0,0,1,3.39,4.62a4.16,4.16,0,0,0-.55,2.07A4.09,4.09,0,0,0,4.66,10.1,4.05,4.05,0,0,1,2.8,9.59v.05a4.1,4.1,0,0,0,3.3,4A3.93,3.93,0,0,1,5,13.81a4.9,4.9,0,0,1-.77-.07,4.11,4.11,0,0,0,3.83,2.84A8.22,8.22,0,0,1,3,18.34a7.93,7.93,0,0,1-1-.06,11.57,11.57,0,0,0,6.29,1.85A11.59,11.59,0,0,0,20,8.45c0-.17,0-.35,0-.53A8.43,8.43,0,0,0,22,5.8Z"/></svg></a></div>
```
Auenc, line 215: Would be nice if you used the `<img` tag to import the SVG here. That way we reduce the number of lines in our HTML and make it a little easier to read :)

styles.css
```diff
@@ -0,0 +1,754 @@
+/* CSS Custom Properties */
+:root {
+  /* border styling */
+  --border-styling: 2px solid #333333;
```
Auenc, line 4: Great use of CSS vars!

styles.css
```diff
@@ -0,0 +1,754 @@
...
+.wrapper {
+    display: grid;
+    grid-template-rows: 0.1fr 0.8fr 0.1fr;
+    gap: 0px 0px;
+    grid-template-areas:
+    "rh-sb"
+    "lh-sb"
+    "main";
+      height: 100vh;
+      /* width: 100vw; */
```
Auenc, line 122: It's always nice if we can remove commented code if it is no longer needed :)

Auenc (approved):
Excellent work! I've made a few comments about improvements that can be made/little bugs you might have :)

PR #5, opened 2021-11-30
Twitter Project - main layout done

New-Arday:


twitter.html
```diff
@@ -0,0 +1,458 @@
...
+  <head>
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Twitter Project</title>
+    <link rel="stylesheet" href="twitter1.css" />
+  </head>
+  <body>
+    <!-- ... icon -->
+    <!-- <svg xmlns="http://www.w3.org/2000/svg" class="h-6 w-6" fill="none" viewBox="0 0 24 24" stroke="currentColor">
```
Auenc, line 12: I think you might have pushed the work you had done when Mike told you to create the PR - but just in case, it's always nice to remove any commented code we no longer need. Just to keep the code base clean and readable :)

twitter.html
```diff
@@ -0,0 +1,458 @@
...
+          <ion-icon name="logo-twitter" class="twitter"></ion-icon>
+          <div class="left-icons-n-list">
+            <div class="left-m-icons">
+              <ul class="list">
+                <li>
+                  <ion-icon name="home" class="house"></ion-icon>
+                </li>
+                <li>
+                  <p>
+                    <svg
```
Auenc, line 28: We can make our code a little easier to read if we import an SVG via the image tag, something similar to this ``

twitter.html
```diff
@@ -0,0 +1,458 @@
...
+            <img
+              src="/images/img/burningcars.jpg"
+              alt="picture"
+              class="pic-profile"
+            />
+          </div>
+
+          <div class="e1-b e2-2">
+            <ion-icon name="language-outline"></ion-icon
+            ><ion-icon name="language-outline"></ion-icon
```
Auenc, line 124: Be careful about having your closing `>` on a newline. It''s a little hard to read. It still works, but difficult for us to read over :)

twitter1.css
```diff
@@ -0,0 +1,168 @@
+* {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
+:root {
+  --right-menu-bg-clr: #f5f8fa;
```
Auenc, line 7: You can combine all of these `:root` variables into one single `:root` :)

Auenc (commented):
Excellent work so far! I've mentioned a few things you can do to make improvements :)

twitter1.css
```diff
@@ -201,8 +192,6 @@ body {
 .clr-grn {
   color: #74b816;
```
Auenc, line 193: Might be a good idea in future that whenever you add a new colour like here `color: #74b816;` you add it to your CSS vars like you have elsewhere :)

index.html
```diff
@@ -0,0 +1,145 @@
+<!DOCTYPE html>
```
Auenc, line 1: I can see that you started again in another file. It would be nice that (if you do this again in future) that you delete the old file. The reason being is we try to keep pull requests as small as possible. Obviously in this case we have asked you to submit all of your work as a pull request, so it's going to be big - however, the `index.html` here isn't needed as you've restarted with the `twitter.html`

twitter.css
```diff
@@ -0,0 +1,84 @@
+ {
```
Auenc, line 1: Same as the above comment, that if you've started again in a different file, it would be a good idea to remove the old files.

twitter1.css
```diff
@@ -0,0 +1,299 @@
+* {
```
Auenc, line 1: Personally I wouldn't apply styles using the `*`. The reason being is that some elements have `margin`/`padding` by default. By removing these properties on every element, you will have to re-add them when you want them. 

This is a personal preference though, so if you prefer to do it that way that is fair enough :)

twitter1.css
```diff
@@ -0,0 +1,299 @@
...
+  width: 100px;
+  height: 100px;
+  padding: 20px;
+  border-radius: 50%;
+}
+.pic-profile-2 {
+  width: 70px;
+  height: 70px;
+  padding: 10px;
+  border-radius: 50%;
```
Auenc, line 160: I can see here (as well as on line 154) that you're setting the border-radius for profile pictures. You could, in future, create one css class for rounding borders, and apply that to both of the profile pictures. For instance
```css
.circle-avatar {
border-radius: 50%;
}
```

That way, you don't have to repeat yourself in the different classes you have.

twitter1.css
```diff
@@ -0,0 +1,299 @@
...
+  display: inline-block;
+  padding-right: 40px;
+  padding-top: 40px;
+  margin-right: 40px;
+}
+.main-pic {
+  width: 600px;
+  height: 600px;
+  justify-content: 2 /-3;
+  border: ;
```
Auenc, line 243: You're not setting a border here, did you mean to delete this line?

twitter.html
```diff
@@ -0,0 +1,333 @@
...
+            />
+          </div>
+        </main>
+      </div>
+
+      <!--  RIGHT MENU -->
+
+      <div class="right-menu">
+        <div class="search-bar">
+          <p class="search-bar">Search Twitter</p>
```
Auenc, line 238: This should be a `<input type="text" />` rather than a `<p>`. The input element will give you a text box that will allow the user to type into to search twitter

twitter.html
```diff
@@ -0,0 +1,333 @@
...
+            <div class="left-m-icons">
+              <ul class="list">
+                <a href="https://twitter.com/">
+                  <li>
+                    <ion-icon name="home" class="house"></ion-icon></li
+                ></a>
+                <a href="#"
+                  ><li>
+                    <p>
+                      <svg
```
Auenc, line 32: You could use the svg with an `` tag using the `src` property to point to the SVG file.

twitter.html
```diff
@@ -0,0 +1,333 @@
...
+            <br />
+            <strong>Lorem ipsum dolor</strong
+            ><span class="gry-font-big">@consectetu 26m</span> <br />Lorem ipsum
+            dolor sit amet consectetur, adipisicing elit. Officia, doloribus? Nu
+            <ion-icon name="stats-chart-outline" class="ylw-clr"></ion-icon>
+            <br /><br />
+            Lorem ipsum dolor sit amet consectetur adipisicing elit. Corrupti
+            nihil consequatur dolorum tenetur fuga
+            <ion-icon name="storefront-outline" class="clr-grn"></ion-icon>
+            <br />
```
Auenc, line 182: We typically prefer not to use `<br />` tags to add vertical spacing to our webpage. Instead, it is better to add vertical spacing with `margin-bottom`.

Auenc:
Overall good work on what you have submitted. I think it would be a good idea to try and test your webpage on different resolutions (using the chrome inspector tool) as on my screen the right-hand menu was very big; as well as some of the components are not displaying as I would expect twitter to. It would have been nice to see some of the hover effects completed, such as your buttons showing the pointer cursor.

PR #8, opened 2021-11-30
twitter review

alsyoufomar:


style.css
```diff
@@ -0,0 +1,577 @@
+* {
+  margin: 0;
```
Auenc, line 2: This is just a personal preference of mine - but I don't like setting `margin:0` using the the `*` selector, because it'll remove the margin from every element on the page. A lot of elements have margin by default, so this means you would have to re-add the margin for those elements.

style.css
```diff
@@ -0,0 +1,577 @@
...
+  font-size: 16px;
+  margin-top: 2px;
+}
+.tweet-nav {
+  width: 180px;
+  margin-left: 0px;
+  position: relative;
+  top: 20px;
+}
+nav img[alt="myself"] {
```
Auenc, line 213: Great use of the selecting an image with a specific `alt` property!

twitter.html
```diff
@@ -0,0 +1,484 @@
...
+      <div class="search">
+        <div class="search-button">
+          <img src="search.png" alt="search-icon" />
+          <input type="text" placeholder=" Search Twitter" />
+        </div>
+      </div>
+      <header>
+        <h2><span class="home1">Home</span></h2>
+        <div class="effect">
+          <svg viewbox="0 0 24 24" aria-hidden="true" class="star">
```
Auenc, line 23: It's a nice practice to use the `` tag when using SVGs as it reduces the amount of lines in our html file and makes it more readable.

Auenc (approved):
Excellent work! I love the sticky "who to follow" section! 

I've made one or two comments that you could improve on with future exercises :)

alsyoufomar:
Thank u so much Lewis I do appreciate it and I will try to consider your
comments next time, However, Mike taught us this trick (margin = 0) and
without it, it will have a gap around the main container.

PR #13, opened 2021-11-30
twitter-submission

nicopicchio:


index.html
```diff
@@ -0,0 +1,383 @@
...
+                            <img id="profile-photo" src="/img/images/profile.jpg" alt="profile photo" width="80px">
+                        </div>
+
+                        <div id="status-right">
+                            <div id="status-right-top">
+                                <h3>What's happening?</h3>                        
+                            </div>
+                            <div id="status-right-bottom">
+                                <div id="status-right-bottom-svg">
+                                    <svg viewbox="0 0 24 24"><g><path d="M19.75 2H4.25C3.01 2 2 3.01 2 4.25v15.5C2 20.99 3.01 22 4.25 22h15.5c1.24 0 2.25-1.01 2.25-2.25V4.25C22 3.01 20.99 2 19.75 2zM4.25 3.5h15.5c.413 0 .75.337.75.75v9.676l-3.858-3.858c-.14-.14-.33-.22-.53-.22h-.003c-.2 0-.393.08-.532.224l-4.317 4.384-1.813-1.806c-.14-.14-.33-.22-.53-.22-.193-.03-.395.08-.535.227L3.5 17.642V4.25c0-.413.337-.75.75-.75zm-.744 16.28l5.418-5.534 6.282 6.254H4.25c-.402 0-.727-.322-.744-.72zm16.244.72h-2.42l-5.007-4.987 3.792-3.85 4.385 4.384v3.703c0 .413-.337.75-.75.75z"></path><circle cx="8.868" cy="8.309" r="1.542"></circle></g></svg>
```
Auenc, line 124: Might be a nice idea to have these SVGs in a file, and import them with something similar to ``. That way we keep the line count of the HTML lower and our code is more readable.

style.css
```diff
@@ -0,0 +1,421 @@
...
+
+.feed-posts-container p.twitter-post {
+    display: block;
+    margin: 8px 0px;
+    width: 60ch;
+    line-height: 1.5;
+}
+
+.post-image {
+    border-radius: 4px !important;
```
Auenc, line 249: Usually it's a bad idea to use `!important`. Typically when you feel you need to use it, you're often applying the css in the wrong place/on the wrong element. 

The reason why we don't like to use `!important` is because if we ever want to overwrite this style, we will use another `!important`.

Auenc (approved):
Great work! I think the only things I could pick out were the `!important`, using the `` tag for SVGs. I think it would have also been nice if all of you "linkable" elements such as profile pictures/names/hashtags and @s had hover effects, but amazing work overall!

index.html
nicopicchio: Yes I have tried to use `` with the `.svg` inside but it was not working.
I think it's because the part with `xmlns` at the start of the svg file was missing but I have only just realised it doing other stuff.

style.css
nicopicchio: Yes I know it's not great but I didn't know how to overcome the issue and I thought it was a good idea using it in view of the size of the clone. On bigger sites it would be different.

PR #16, opened 2021-11-30
Twitter Exercice - Tweak

dan1212111:


mytwitter.html
```diff
@@ -0,0 +1,67 @@
...
+            <ul>
+                <li>logo 1</li>
+                <li>logo2 </li>
+                <li>logo</li>
+                <li>logo</li>
+                <li>logo</li>
+                <li>logo</li>
+                <li>logo</li>
+                <li>logo</li>
+
```
phillipdunn, line 25: unnecessary space

safeguard/safeguard.css
```diff
@@ -0,0 +1,155 @@
+* {
+    box-sizing: border-box;
+    margin: 0%;
+    padding: 0%;
+
```
phillipdunn, line 5: unnecessary space

safeguard/safeguard.css
```diff
@@ -0,0 +1,155 @@
...
+}
+
+#layout {
+    height: 100vh;
+    display: grid;
+    grid-template:
+
```
phillipdunn, line 12: unnecessary space.. i think you get the jist :-)

phillipdunn:
Good effort here, it looks like you have pushed a few half done files that you probably didnt want to, make sure you are only commiting/pushing the code that you actully want to submit in the directory, ommit or get rid of the stuff you dont need. Also a few comments about your code presentation.

dan1212111:
Hi Phil,

Would it be possible to have a more in-depth review of my Twitter exercise? For instance how efficient have I been in getting around the problem? Any unnecessary elements in my code besides the spacings? Was my syntax use appropriate? I need more information to improve upon.

dan1212111:
The most recent file is the correct file.

styles.css
```diff
@@ -0,0 +1,1240 @@
+* {
+    box-sizing: border-box;
+    margin: 0%;
```
Auenc, line 3: This is a personal preference of mine, but I prefer not to set the `margin: 0;` and `padding:0` using the `*` selector. The reason being is that it will remove the margin/padding from every element on the page. 

Personally, I prefer to remove the margin/padding from the html/body elements. That way, you get rid of the initial padding/margin from the page, but also keep the default margin/padding on elements such as the `p` tag (there are many other elements with default padding/margin, but p is just an example)

styles.css
```diff
@@ -0,0 +1,1240 @@
...
+.main-right {
+    grid-area: main-right;
+    min-width: 200px;
+    border-radius: 25px;
+    margin-top: 2mm;
+    margin-right: 2mm;
+    margin-bottom: 2mm;
+    margin-left: 2mm;
+    background-color: rgb(247 249 249);
+    );
```
Auenc, line 978: This is causing an error when trying to view the submission in code sandbox. You might find the solution is working on your machine (browsers can sometimes ignore errors like these and still render the page) but a syntax error like this would cause your webpage to not render when we come to the later modules, so just be careful with it! If visual studio code is showing a filename in red it means there is an error in the file. Typically you'll be able to see where the error is on the right hand side, it should show a small red line next to the scroll bar

tweaktweak.html
```diff
@@ -0,0 +1,371 @@
...
+    <link rel="stylesheet" href="styles.css">
+    <title>🦅tweak</title>
+</head>
+
+<body>
+    <div id="layout">
+        <div class="side"></div>
+        <div class="left-header"><img src="icons/logo-blue.svg"></div>
+        <div class="list-logo">
+            <div class="list-logo1">
```
Auenc, line 16: It's a good idea to have descriptive class names. For instance, what is the difference between `list-logo` and `list-logo1`? Looking at your code (I could be wrong) it looks like `list-logo` might have been better named `list-logo-container`. Or something similar to that :thinking:

tweaktweak.html
```diff
@@ -0,0 +1,371 @@
...
+                    </li>
+                    <li>
+                        <div id="more11">
+                            <img style="vertical-align:middle" src="icons/more-blue.svg">
+                            <span>More</span>
+                        </div>
+                    </li>
+                </ul>
+            </div>
+            <div class="blank-list">
```
Auenc, line 74: I'm not really sure what this list is for? Is it to add spacing to the left hand side of your menu? You might have been able to achieve the same effect using `margin-right` on one of your elements within the left hand menu. Or even padding-left on your list

tweaktweak.html
```diff
@@ -0,0 +1,371 @@
...
+
+
+        <div class="tweak">Tweak</div>
+        <div class="bottom-left">
+            <div id="bottom-profile"><img src="dimages/profile.png"></div>
+            <div id="bottom-info">
+                <h4>Daniel McCarthy</h4>
+                <p>@Dan1212</p>
+            </div>
+            <div id="three-dots"><svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"
```
Auenc, line 120: It's a good practice to include SVGs in an `` tag such as ``. This way we reduce the number of lines in our html and make it more readable :)

Auenc: Ah I can see that you've done that lower down ``.

tweaktweak.html
```diff
@@ -0,0 +1,371 @@
...
+                <div id="media"><img src="icons/media.svg"></div>
+                <div id="gif"><img src="icons/gif.svg"></div>
+                <div id="poll"><img src="icons/poll.svg"></div>
+                <div id="emoji"><img src="icons/emoji.svg"></div>
+                <div id="schedule"><img src="icons/schedule.svg"></div>
+            </div>
+            <div id="small-tweak">Tweak</div>
+        </div>
+        <div class="main">
+            <div class="tweak1">
```
Auenc, line 145: I think this goes back to a previous comment, but it's good practice to have descriptive class names. So how does `tweak` differ from `tweak1`, and perhaps try to capture that difference in the class name.

styles.css
```diff
@@ -0,0 +1,1240 @@
...
+    grid-template:
+
+        "side    left-header   left-header     main-header     right-header   side2"50px "side    list-logo     list-logo    middle-main     main-right     side2"15% "side    list-logo    list-logo    main            main-right     side2"1fr "side     tweak         tweak    main              main-right    side2"75px "side     bottom-left  bottom-left     main            bottom-right  side2"70px/ 1fr 0.8fr 1.3fr 6fr 3fr 0.8fr;
+
+}
+
+.side {
+    grid-area: side;
+    background-color: white;
+    ;
```
Auenc, line 20: This `;` isn't needed here

styles.css
```diff
@@ -0,0 +1,1240 @@
...
+    width: 100%;
+    border-radius: 60%;
+}
+
+.list-logo {
+    grid-area: list-logo;
+    display: grid;
+    grid-template-columns: repeat(2, auto);
+    margin-top: 4mm;
+    margin-left: 4mm;
```
Auenc, line 43: It's not really recommended to use `mm` as a unit for creating webpages. It's more useful when creating styles that are specifically designed to be printed as that way the real world measurements relate to the paper being printed.

styles.css
```diff
@@ -0,0 +1,1240 @@
...
+
+    width: 110px;
+    padding-left: 4mm;
+    border-radius: 4cm;
+    margin-top: 2mm;
+}
+
+
+#home:hover {
+    background-color: rgb(231 231 232);
```
Auenc, line 129: Suggested improvement: It looks like you're using this same colour a lot. You could use CSS Vars to create a variable, and reuse it elsewhere in your styles. You can see an example of doing this here: https://www.w3schools.com/css/css3_variables.asp

styles.css
```diff
@@ -0,0 +1,1240 @@
...
+    background-color: rgb(22 124 192);
+}
+
+.bottom-left {
+    grid-area: bottom-left;
+    margin-bottom: 5mm;
+    padding-top: 2mm;
+    background-color: white;
+    display: grid;
+    grid-template: "pic  info     dots"1fr/ 20% 60% 20%;
```
Auenc, line 193: This might be an issue with how github is displaying the code, but it looks like you might have a tab in your `grid-template` in-between info and dots. This shouldn't effect how your grid works, just make's it a little less readable

styles.css
```diff
@@ -0,0 +1,1240 @@
...
+.middle-main {
+    grid-area: middle-main;
+    background-color: white;
+    border-top: 2px solid rgb(204 215 221);
+    border-left: 2px solid rgb(204 215 221);
+    border-right: 2px solid rgb(204 215 221);
+    padding-bottom: 1mm;
+    min-width: 300px;
+    display: grid;
+    grid-template: "pic  main    main"2fr "pic   icons   tweak"1fr/ 10% 1fr 66px;
```
Auenc, line 288: This is also a note on readability, but it's easier to read the `grid-template` if you place each row on a new line.

styles.css
```diff
@@ -0,0 +1,1240 @@
...
+    border-radius: 100px;
+}
+
+#emoji:hover {
+    box-shadow: 12px 12px 2px 1px rgba(0, 0, 255, .2);
+    border-radius: 100px;
+}
+
+#schedule:hover {
+    box-shadow: 12px 12px 2px 1px rgba(0, 0, 255, .2);
```
Auenc, line 350: A lot of these styles look to have the same rules (schedule, emoji, poll, gif, media). It would be better to do this as one style and apply it to each of the elements. That way you don't have to repeat the styles you've already written.

styles.css
```diff
@@ -0,0 +1,1240 @@
...
+    margin-right: 2.6cm;
+
+}
+
+#tweak-body1 span:hover {
+    text-decoration: underline;
+}
+
+#tweak-body4 span:hover {
+    text-decoration: underline;
```
Auenc, line 893: This is another example of repeated styles. You could create one class called `hover-underline` or something similar, add the text decoration when it's hovered: 
```css
.hover-underline:hover {
  text-decoration: underline;
}
``` 

Then apply it to each of the elements that you want to be underlined when they're hovered. That way you can achieve the same result with only one class.

Auenc:
Overall excellent work! There are a few comments I've made that you can use going forward in future exercises that I think would save you some time (and line count!). Some additional things I picked up while viewing the page - I've noticed you've added hover effects for most of the things I'd expect to have them, however, it would have been nice to add the `cursor: pointer` rule when hovering them (for things that should be clickable). This small rule gives a lot more "feedback" to the user that the element is clickable. 

You did a great job overall at recreating twitter! I would say that the right hand side looks a little off on my screen; it looks like it has a bit too much spacing in some places.

I think the most important thing to practice in future exercises is that css classes are reusable. So if you find yourself writing the same style for several different elements, make it one style and apply that style to the elements :)

dan1212111:
This feedback is great and extremely helpful,  thanks a lot!

PR #17, opened 2021-11-30
Twitter Submission

7LGnGZkY:


index.html
```diff
@@ -0,0 +1,78 @@
...
+            <input type="text" placeholder="Search Twitter" class="search-twitter">
+            <br><br>
+            <div class="trending">What's happening</div>
+            <div class="trending">2</div>
+            <div class="trending">3</div>
+            <div class="trending">4</div>
+            <div class="trending">5</div>
+        </div>     
+        
+        <!-- <div class ="happening">
```
phillipdunn, line 65: avoid committing commented code, if you don't need it, delete it

style.css
```diff
@@ -0,0 +1,162 @@
...
+}
+
+.layout {
+    display:grid;
+    grid-template-columns: 0.8fr 2fr 1fr;
+    grid-gap: 1em;
+    grid-auto-rows: minmax(100px, auto);
+    padding-inline: 10em;
+    
+    /* border: #333 3px solid; */
```
phillipdunn, line 25: same as above

phillipdunn:
The basic layout is there and well laid out, but a decent part of the task was to add images, icons, and tweets. which are not in there. Tidy looking code, but not quite there.

index.html
```diff
@@ -0,0 +1,244 @@
...
+</head>
+<body>
+    <div class="layout">
+        <div class="left">
+            <div> <a class="logo" href="https://www.twitter.com">
+                <img alt="Twiter" src="images/logo.svg" height="27px">
+                </a> </div>
+            <div></div>
+            <div>
+                <svg viewBox="0 0 24 24" aria-hidden="true" height ="23px" class="r-18jsvk2 r-4qtqp9 r-yyyyoo r-lwhw9o r-dnmrzs r-bnwqim r-1plcrui r-lrvibr"><g><path d="M22.58 7.35L12.475 1.897c-.297-.16-.654-.16-.95 0L1.425 7.35c-.486.264-.667.87-.405 1.356.18.335.525.525.88.525.16 0 .324-.038.475-.12l.734-.396 1.59 11.25c.216 1.214 1.31 2.062 2.66 2.062h9.282c1.35 0 2.444-.848 2.662-2.088l1.588-11.225.737.398c.485.263 1.092.082 1.354-.404.263-.486.08-1.093-.404-1.355zM12 15.435c-1.795 0-3.25-1.455-3.25-3.25s1.455-3.25 3.25-3.25 3.25 1.455 3.25 3.25-1.455 3.25-3.25 3.25z"></path></g></svg>
```
phillipdunn, line 18: svg's should live in thier own file and not in the html, use should use an  tag and then point the src to the path of the svg

index.html
```diff
@@ -0,0 +1,244 @@
...
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="stylesheet" href="style.css">
+    <title>Home / Twitter</title>
+</head>
+<body>
+    <div class="layout">
+        <div class="left">
+            <div> <a class="logo" href="https://www.twitter.com">
```
phillipdunn, line 13: <a> you noramlly want to open in a different page/tab so use target="_blank" inside the tag to do this

phillipdunn:
Looking much better, really good effort here. A few comments to help improve, not much else to add. Well done.

PR #19, opened 2021-11-30
Twitter Assignment

BAZBOT123:


index.html
```diff
@@ -0,0 +1,538 @@
...
+                        <div class="disappear">
+                            <h2><a href="Batman.html">More</a></h2>
+                        </div>
+                    </div>
+
+                    <div button class="box9">
+                        <a href="Batman.html" class="lefttweet">Tweet</a>
+                    </div>
+
+
```
phillipdunn, line 78: avoid gaps like this

index.html
```diff
@@ -0,0 +1,538 @@
...
+        <!-- MIDDLE COLUMN -->
+
+
+        <!-- TOP BOX /MID COLUMN -->
+        <div class="main-col">
+            <div class="boxone">
+                <div class="homeone">
+                    <div class="top-home">Home</div>
+                    <div class="top-tweets">
+                        <svg viewBox="0 0 24 24">
```
phillipdunn, line 111: svg's should not be inside the html, but be in a seperate file for all your svg's. You should use an  tag and then point to the svg file with the src.

index.html
```diff
@@ -0,0 +1,538 @@
...
+                        <div class="tweetbut">
+                            <a href="Batman.html" class="btn">Tweet</a>
+                        </div>
+                    </div>
+
+
+
+
+
+                    <!--<ul class="posts">
```
phillipdunn, line 209: avoid committing commented out code,

styles.css
```diff
@@ -0,0 +1,646 @@
...
+    display: grid;
+}
+
+
+.layout {
+    background-color: white;
+    display: grid;
+    height: 105vh;
+    grid-template: 
+        "left-col        main-col        right-col"     1fr/
```
phillipdunn, line 16: no need to space things out so much try to keep it reasonably spaced e.g. "left-col  main-col  right-col" etc

styles.css
```diff
@@ -0,0 +1,646 @@
...
+    margin-top: 10px;
+}
+
+.profile-pic,
+.deets {
+    display: grid;
+    place-items: center;
+}
+
+
```
phillipdunn, line 139: quite a few gaps again in this file that are not necessary

phillipdunn:
Overall looks really good, just some tidying up and in terms of code being tidy and easy to read

PR #21, opened 2022-02-18
Update README.md

Creatorrr1:


twitter.css
```diff
@@ -0,0 +1,279 @@
...
+  margin-bottom: 40px;
+  margin-right: 20px;
+  cursor: pointer;
+  
+}
+
+.no-decoration {
+  text-decoration: none;
+  list-style: none;
+}
```
Creatorrr1, line 286: Creating a whole class for list decoration is really useful when you got more than 1 list. Nice one.

twitter.css
```diff
@@ -0,0 +1,279 @@
...
+  font-size: 14px !important;
+  color: var(--twitter-color);
+  margin-right: 5px;
+}
+
+.post__header-Special {
+  font-weight: 600;
+  font-size: 12px;
+  color: gray;
+}
```
Creatorrr1, line 238: Adding a greyish color to create a transparent like username to create a difference in tone from other text, worked well.

twitter.css
```diff
@@ -0,0 +1,279 @@
...
+  padding: 20px;
+}
+
+.side-bar-menu :hover {
+  background-color: var(--twitter-background);
+  border-radius: 30px;
+  color: var(--twitter-color);
+  transition: color 100ms ease-out;
+}
+
```
Creatorrr1, line 109: Creating a variable for a colour so that you can always refer back to it was really useful and efficient.

index.html
```diff
@@ -0,0 +1,213 @@
...
+            </ul>
+          </nav>
+          <button class="side-bar-button__tweet">Tweet</button>
+        </aside>
+      </header>
+      <main>
+        <div class="main-box">
+          <h3>Home</h3>
+        </div>
+          <div class="  profile-feed-pic main-box">
```
Alex90Jennings: did you mean to open a div "main box" then close it immediately with only a h3 inside? The next div "profile-feed-pic" is indented, which suggests that it wants to be the child of the class above. Should the divs be siblings or parent-child?

twitter.css
```diff
@@ -0,0 +1,290 @@
...
+.main-header {
+  width: 275px;
+  height: 100vh;
+}
+
+.responsive-rows {
+  display: grid;
+  grid-auto-rows: repeat(auto-fit, minmax(800px, 1fr));
+}
```
Alex90Jennings, line 48: nice grouping of CSS grids, makes it easy to read and edit later if needed

twitter.css
```diff
@@ -0,0 +1,290 @@
...
+
+.main-box {
+  border: 2px solid var(--twitter-background);
+  margin: 3px;
+  min-height: 56px;
+  position: sticky;
+}
+
+.post-box {
+  border: 2px solid var(--twitter-background);
```
Alex90Jennings: a bit of repetition of border: 2px solid var(--twitter-background); perhaps a new class could have saved some time, as this border may be useful later

twitter.css
```diff
@@ -0,0 +1,290 @@
...
+  color: black;
+}
+
+.menu-list img {
+   font-weight: 800;
+   font-size: 20px;
+   margin-right: 10px;
+   padding: 30px;
+   padding-bottom: 0px;
+  }
```
Alex90Jennings, line 181: font attributes when targeting an img?

twitter.css
```diff
@@ -0,0 +1,290 @@
...
+}
+
+.post__headerDescription {
+  margin-bottom: 10px;
+  font-size: 15px;
+}
+
+.post__footer {
+  display: flex;
+  justify-content: space-between;
```
Alex90Jennings, line 247: i've never seen "space-between" would be interested to know what it does

twitter.css
```diff
@@ -0,0 +1,290 @@
...
+  font-weight: 900;
+  width: 80px;
+  height: 40px;
+  margin-bottom: 40px;
+  margin-right: 20px;
+  cursor: pointer;
+  
+}
+
+.no-decoration {
```
Alex90Jennings, line 283: really like the use of a no-decoration class. could you have used this on the a tag above (line 103)

index.html
```diff
@@ -0,0 +1,213 @@
...
+        <div class="News">
+          <div class="News__input">
+            <span class="material-icons News__searchIcon"> search </span>
+            <input type="text" placeholder="Search Twitter" />
+          <!-- <News></News> -->
+          <div class="News__widgetContainer">
+            <h2>What's happening?</h2>
+          </div>
+        <!-- DOES IT BREAK THE NATURAL FLOW? WILL THE CONTENT PROVIDE THE HEIGHT? -->
+      </aside>
```
Alex90Jennings: nice use of asides to avoid a div soup

index.html
```diff
@@ -0,0 +1,213 @@
...
+            <img
+              class="twitter-logo"
+              src="https://img.search.brave.com/jaw-9w6G39GIIxdKdqgSGOIdknwh344YLISmmJE_2rw/rs:fit:1200:1200:1/g:ce/aHR0cHM6Ly9sb2dv/cy1kb3dubG9hZC5j/b20vd3AtY29udGVu/dC91cGxvYWRzLzIw/MTYvMDIvVHdpdHRl/cl9Mb2dvX25ldy5w/bmc"
+              alt="Twitter Logo"
+            />
+          </div>
+          <nav>
+            <ul class="menu-list">
+
+              <li class="active">
```
Alex90Jennings, line 35: nice use of an active class, makes it really quick and simple to add or remove this property

twitter.css
Alex90Jennings: agree

twitter.css
```diff
@@ -0,0 +1,290 @@
...
+
+span img {
+  height: 40px;
+  border-radius: 30px;
+  object-fit: contain;
+}
+
+/* POST */
+.post__badge {
+  font-size: 14px !important;
```
Alex90Jennings, line 229: what does !important do?

index.html
```diff
@@ -0,0 +1,213 @@
...
+                  <img
+                    src="https://img.search.brave.com/khlf_hjYm7Odq9bMccxhKQLM8IrcZC5b8BhDEKo3-c0/rs:fit:1200:940:1/g:ce/aHR0cHM6Ly9pMC53/cC5jb20vdmlyYWwu/b2xhc21lZGlhdHYu/Y29tL3dwLWNvbnRl/bnQvdXBsb2Fkcy8y/MDIyLzAyL1dBVENI/LUFtaXItS2hhbi12/cy1LZWxsLUJyb29r/LUZ1bGwtRmlnaHQt/U3RyZWFtaW5nLU9u/bGluZS5qcGc_em9v/bT0yJnJlc2l6ZT03/ODAlMkM0NzA"
+                    alt=""
+                  />
+                  <div class="post__footer">
+                    <span class="material-icons"> repeat </span>
+                    <span class="material-icons"> favorite_border </span>
+                    <span class="material-icons"> publish </span>
+                  </div>
+                </div>
```
Alex90Jennings, line 171: I don't see where the div class="post" closes? for me, in the tweets, there is one too few div closing attributes (3 open, 2 closes)

twitter.css
qruz-hq: This is actually really useful ! I was using it on the `ul` tag directly but it wouldn't allow to have actual lists in the page. Thanks

index.html
```diff
@@ -164,6 +197,14 @@ <h3>
         <!-- DOES IT BREAK THE NATURAL FLOW? WILL THE CONTENT PROVIDE THE HEIGHT? -->
       </main>
       <aside class="resources-sidebar">
```
qruz-hq: Nice use of the aside & main tag we can quickly identify the different sections of the page

twitter.css
```diff
@@ -131,24 +143,59 @@ a {
...
+}
+
+.feed-menu-icon {
+  width: 25px;
+  height: auto;
+}
 
 menu-list-link {
 
 }
```
qruz-hq: This could be removed

twitter.css
```diff
@@ -0,0 +1,290 @@
...
+  /* background-color: var(--twitter-background);
+  border-radius: 40px; */
+  color: var(--twitter-color);
+  transition: color 100ms ease-out;
+
+}
+
+.feed-menu-icon {
+  width: 25px;
+  height: auto;
```
qruz-hq: I think adding `height: auto;` is a bit redundant because that's the initial state of the attribute

index.html
Creatorrr1: The code wasn't finalised, the feed icons (e.g the gif ) was imperfect and this snippet was incomplete and was to be used later when i get a hold of the sizing for the icons.

twitter.css
Creatorrr1: I agree but there was only 2 boxes i really needed for the main feed and right-side so adding extra class wouldn't make sense except for longer projects.

twitter.css
Creatorrr1: Nice catch didn't see that, copied a pasted some things i needed, forgot to remove the font attributes.

twitter.css
Creatorrr1: Yes this was just something I saw someone do before so when i implemented it, it made everything easier.

index.html
Creatorrr1: Thanks your right I'll get that sorted.

twitter.css
Creatorrr1: This ignores any specificity rules applied beforehand and makes this attribute the default instead.

PR #22, opened 2022-02-18
Paulo Barata

PBara7a:


styles/style.css
```diff
@@ -161,6 +161,37 @@ main .feed .tag {
...
+main .feed .retweet svg:hover {
+    fill: #47bb2a;
+}
+
+main .feed .like:hover {
+    background-color: #f918801a;
+}
+
+main .feed .like svg:hover {
+    fill: #f91880;
```
PBara7a: I am not happy with the behaviour of these svg:hover. It only changes the color of the svg if the mouse pointer is exactly over a vector. Any suggestion on how to change the color of the svg when the parent button is hovered?

styles/style.css
```diff
@@ -56,16 +56,20 @@ a:active {
...
+
+    overflow-y: scroll;
+    /* Hide scrollbar */
+    -ms-overflow-style: none;  /* IE and Edge */
+    scrollbar-width: none;  /* Firefox */
+}
+
+main::-webkit-scrollbar {
+    /* Hides scrollbar for Chrome, Safari and Opera */
+    display: none;
```
PBara7a, line 85: It was interesting to find that I need different lines of code, each one implementing the same behaviour on different browsers. Need to find if there is a shorthand to hide scrollbars on every browser.

styles/style.css
```diff
@@ -69,6 +69,11 @@ button:hover {
     padding: 16px 0px;
 }
 
+.left-menu nav a:focus {
+    background-color: var(--third-color);
+    outline: none;
+}
```
PBara7a, line 68: Important feature to style selected elements when the user is not navigating with a mouse.

styles/style.css
```diff
@@ -300,21 +330,39 @@ main .bottom-nav a {
...
+        align-content: center;
+        justify-content: center;
+    }
+    .left-menu .two-column-grid_expand-two {
+        grid-template-columns: auto;
+    }
+
+    .left-menu-link-text {
         display: none;
     }
```
PBara7a: Reminder: the size of elements with `display: none` is not taken into account.

index.html
```diff
@@ -12,34 +12,47 @@
     <div class="center-wrapper three-column-grid__expand-two">
         <aside class="left-menu asides-padding">
             <ul>
-                <img class="margin-bottom-md" src="/icons/twitter.png" alt="Twitter icon.">
```
AlexDjangoX: I never thought to add the img-tag on line 15, between the ul-tag and li-tag to keep it in the ul container - better for alignment and styling?

index.html
```diff
@@ -12,34 +12,47 @@
...
+                        <h2>Explore</h2></a>
                 </li>
-                <li class="two-column-grid_expand-two gap-bg align-center">
-                    <img src="/icons/bell.png" alt="Bell icon.">
-                    <h2><a href="#">Notifications</a></h2>
+                <li>
+                    <a class="two-column-grid_expand-two gap-md align-center" href="">
+                        <svg fill="currentColor" height="26px" width="26px"><g><path d="M21.697 16.468c-.02-.016-2.14-1.64-2.103-6.03.02-2.532-.812-4.782-2.347-6.335C15.872 2.71 14.01 1.94 12.005 1.93h-.013c-2.004.01-3.866.78-5.242 2.174-1.534 1.553-2.368 3.802-2.346 6.334.037 4.33-2.02 5.967-2.102 6.03-.26.193-.366.53-.265.838.102.308.39.515.712.515h4.92c.102 2.31 1.997 4.16 4.33 4.16s4.226-1.85 4.327-4.16h4.922c.322 0 .61-.206.71-.514.103-.307-.003-.645-.263-.838zM12 20.478c-1.505 0-2.73-1.177-2.828-2.658h5.656c-.1 1.48-1.323 2.66-2.828 2.66zM4.38 16.32c.74-1.132 1.548-3.028 1.524-5.896-.018-2.16.644-3.982 1.913-5.267C8.91 4.05 10.397 3.437 12 3.43c1.603.008 3.087.62 4.18 1.728 1.27 1.285 1.933 3.106 1.915 5.267-.024 2.868.785 4.765 1.525 5.896H4.38z"></path></g></svg>
```
AlexDjangoX: Steven put the svg and p (you have used h2) in their own divs whithin the a-tag. It might sort your alignment problem?

index.html
```diff
@@ -148,6 +148,45 @@ <h1>Home</h1>
                     </li>
                 </ul>
             </section>
+            <nav class="bottom-nav">
+                <ul class="grid-auto__columns">
```
AlexDjangoX, line 764: Very good. I like the way you have another class=""bottom-nav" for responsiveness

styles/style.css
AlexDjangoX: You get an add-on that takes your css and adds code for different browsers. When you use SASS which get's compiled, it automatically adds code for other browsers. So must be an add on for normal css

styles/style.css
```diff
@@ -0,0 +1,435 @@
...
+}
+
+.search-bar {
+    border-radius: 35px;
+    background-color: var(--third-color);
+
+    padding: 0px 24px;
+}
+
+/* UTILITY STYLES */
```
Dan2024, line 339: I like the use of these utility styles

styles/layouts.css
```diff
@@ -0,0 +1,35 @@
...
+
+.two-column-grid_expand-two {
+    display: grid;
+    grid-template-columns: auto 1fr;
+}
+
+.three-column-grid__expand-two {
+    display: grid;
+    grid-template-columns: auto 1fr auto;
+}
```
Tibor22, line 35: Good ideal to separate the layout

styles/style.css
```diff
@@ -0,0 +1,435 @@
...
+    padding: 12px 36px;
+    
+    background-color: var(--main-color);
+
+    border-radius: 35px;
+}
+
+#tweet-button:hover {
+    color: #ffffff;
+    background-color: #1ea1f398;
```
Dan2024, line 255: could this colour have been made a custom variable too since it gets used a few more times further down the document?  `var(--clr-hover)` perhaps?

styles/style.css
Tibor22: Thanks for that information :D

index.html
```diff
@@ -0,0 +1,841 @@
...
+<body>
+    <div class="center-wrapper three-column-grid__expand-two">
+        <aside class="left-menu asides-padding">
+            <nav>
+                <ul>
+                    <img id="logo" class="margin-bottom-md" src="/icons/twitter.png" alt="Twitter icon.">
+                    <li>
+                        <a class="two-column-grid_expand-two gap-md align-center" href="#">
+                            <svg fill="currentColor" height="26px" width="26px" alt="Home icon.">
+                                <g>
```
Dan2024, line 21: whats the meaning of the g tag?

styles/style.css
Tibor22: maybe `main .feed .like:hover svg`

styles/style.css
Tibor22, line 255: How did you get this background-color?

styles/style.css
PBara7a: That works exactly as I wanted, thank you for this solution!

styles/style.css
PBara7a: Thank you. I'll defenitely have a read about SASS in the near future.

index.html
PBara7a: The idea behind keeping the logo image inside the ul at first was to treat it like a "home" button linking to the main page, so it would be part of the navigation like the others. But since there is an actual "home" button already I ended up not implementing that on the logo and left it like this for simplicity when aligning elements.

index.html
PBara7a: That line was the one that was taken out, but I implemented a very similar solution with the svg and a div holding the text. The problem with the alignment is that when I increase the size of the svg, its container takes more space but the drawing stays the same. This forces sibling elements to move or the line height to increase while the icon I am trying to make bigger is not affected at all.

styles/style.css
PBara7a: Yes, that is a good idea. Thank you!

index.html
PBara7a: The `<g>` is a grouping tag, its meant to group its children. So, multiple "drawings" can be grouped to make one picture/icon.

styles/style.css
PBara7a: The `--main-color: #1ea1f3;` was taken from twitter using a chrome extension called "Eye Dropper". This specific color used here is just that main color with some transparency added to it.

PR #23, opened 2022-02-18
Chris Hyde

Hoolio333:


style.css
```diff
@@ -0,0 +1,5 @@
+/* Reset  */
+
+body {
+  background-color: aqua;
+}
```
Hoolio333: Quick, clear & easy way to show Style Sheet is linked.

index.html
```diff
@@ -8,6 +8,62 @@
...
+      </div>
+
+      <div class="right-column">
+        <div class="whats-happening"></div>
+      </div>
+    </main>
+
+    <!-- Footer Section - Not needed -->
+
+    <!-- <Footer></Footer> -->
```
Hoolio333: Footer not needed, therefore not needed to include. Could clutter code.  Deleted at the end, but probably not needed at all.

index.html
```diff
@@ -8,6 +8,62 @@
...
+          </div>
+
+          <div class="main-feed"></div>
+        </div>
+      </div>
+
+      <div class="right-column">
+        <div class="whats-happening"></div>
+      </div>
+    </main>
```
Hoolio333: Clear, reasonable layout.

index.html
```diff
@@ -12,16 +12,14 @@
...
-          <div class="home"></div>
-          <div class="stars-icon"></div>
+        <div class="home-section home-grid">
+          <div class="home">Home</div>
+          <div class="stars-icon">Stars</div>
         </div>
 
-        <div class="search"></div>
+        <div class="search">Search</div>
       </div>
```
Hoolio333: Clear git message, showing reviewer exactly what changes have been made.

index.html
```diff
@@ -300,9 +300,5 @@ <h3>Relationship Advice</h3>
         </div>
       </div>
     </main>
-
-    <!-- Footer Section - Not needed -->
-
-    <!-- <Footer></Footer> -->
```
Hoolio333: Getting rid of code as not needed.  Makes code cleaner & easier to review.

style.css
```diff
@@ -222,9 +222,12 @@ a {
   padding-left: 10ps;
 }
 
-  grid-template: 
-    "profile-pic-bottom username three-dots-icon" a /
+.profile-info-grid {
+    display: grid;
+  grid-template : 
+    "profile-pic-bottom username three-dots-icon" auto /
     auto auto auto;
```
Hoolio333, line 229: Grid doesn't work.  Was added late at the end, needed more time to complete.  Should be inline in column not shown as a row.

style.css
```diff
@@ -53,7 +54,35 @@ body {
...
   grid-area: search;
   padding-right: 100px;
-  /* border: 2px solid rgb(29, 26, 226); */
+  border: 2px solid rgb(226, 225, 225);
+  background-color: rgb(226, 225, 225);
+  border-radius: 30%;
+  text-align: center;
+  text-decoration: black;
+  justify-content: center;
+}
```
Hoolio333: Search doesn't look right.  Work needed to get colour better & search looking more accurate.

style.css
```diff
@@ -122,6 +122,11 @@ body {
   text-align: center;
   padding-top: 10px;
   padding-right: 10px;
+  background-color: white;
+}
+
+.home-button:hover {
+  background-color: rgb(4, 134, 240);
```
Hoolio333: Hover works well & looks good - Use in future. Try to add for images & other links in future.

style.css
Hoolio333: Borders could have been added to sections to distinguish between each other.

style.css
```diff
@@ -0,0 +1,375 @@
...
+
+body {
+  min-height: 100vh;
+  margin-left: 40px;
+  margin-right: 40px;
+  scroll-behavior: smooth;
+}
+
+/* Header Section  */
+
```
PBara7a, line 18: I like the way you organized your stylesheet by sections. It looks tidy!

index.html
```diff
@@ -0,0 +1,304 @@
...
+    <title>Twitter Challenge</title>
+    <link rel="stylesheet" href="style.css" />
+  </head>
+  <body>
+    <!-- Header Section  -->
+
+    <header>
+      <div class="header-grid">
+        <div class="twitter-logo">
+          <img src="/Pictures/icons/Twitter logo.png" alt="" />
```
PBara7a: There are alt attributes on all the images apart from this one, I think that's great.

index.html
```diff
@@ -0,0 +1,304 @@
...
+              <img
+                src="./Pictures/icons/more.png"
+                alt="Three dots"
+                height="30px"
+                width="30px"
+              />
+            </div>
+          </div>
+        </div>
+
```
PBara7a, line 144: The spacing on your html makes it easy to read.

index.html
```diff
@@ -0,0 +1,304 @@
...
+        </div>
+        <div class="space1"></div>
+      </div>
+    </header>
+
+    <!-- Main Section  -->
+
+    <main>
+      <div class="main-section-grid">
+        <div class="left-menu">
```
PBara7a, line 40: Some divs could have a more appropriate tag. I suggest an `<aside>` tag here for example.

style.css
```diff
@@ -0,0 +1,375 @@
...
+.header-grid {
+  display: grid;
+  grid-template:
+    "twitter-logo home-grid search" auto /
+    auto 1fr auto;
+  gap: 5px;
+}
```
PBara7a, line 25: Since we're creating a grid with a single column here I would suggest a more concise way of doing it:
`.header-grid {
  display: grid;
  grid-template-columns: auto 1fr auto;
  
  gap: 5px;
}`
This way there is no need to assign grid-areas later.

style.css
```diff
@@ -0,0 +1,375 @@
...
+  grid-auto-flow: column;
+  font-weight: bolder;
+}
+
+.left-menu {
+  grid-area: left-menu;
+}
+
+.home-button {
+  text-align: center;
```
PBara7a, line 110: Some attributes are repeated often. Maybe you could create a "utility class" containing that attribute and apply to any element you want instead.

style.css
```diff
@@ -0,0 +1,375 @@
...
+.main-section-grid {
+  display: grid;
+  grid-template:
+    "left-menu middle-main right-column" auto /
+    auto 1fr auto;
+}
+
+.left-main-grid {
+  display: grid;
+  grid-auto-rows: auto;
```
Tibor22, line 96: I think the default value is auto so you don't need this line.

style.css
```diff
@@ -0,0 +1,375 @@
...
+
+.profile-button {
+  text-align: center;
+  padding-top: 10px;
+  padding-right: 10px;
+  background-color: white;
+}
+
+.profile-button:hover {
+  background-color: rgb(4, 134, 240);
```
Tibor22, line 188: try to make this color reusable

index.html
```diff
@@ -28,20 +28,57 @@
...
+          <div class="explore">
+            <div class="explore-image">image</div>
+            <div class="explore-button">explore</div>
+          </div>
+
+          <div class="notifications">
+            Notifications
+            <div class="notifications-image"></div>
+            <div class="notifications-button"></div>
+          </div>
```
AlexDjangoX: There are semantically more accurate tags for navigation, which means people using assisted technology like screen readers can tab through your navigation menu. Something like below? If you have `<div>` around your image and text within the `<a>,` they both serve as links

`<nav class="s">
            <ul class="">
                <a href="#">
                    <div class="frame">
                        home icon
                    </div>
                    <div class="text"> Home</div>
                </a>
            </ul>
        </nav>`

index.html
```diff
@@ -0,0 +1,304 @@
...
+              <p>
+                Tweet Example <br />
+                Lorem ipsum dolor sit amet consectetur adipisicing elit. Quae,
+                ea! Lorem ipsum dolor sit amet consectetur adipisicing elit.
+                Quae, ea!
+              </p>
+            </div>
+
+            <div class="main-two-column-grid">
+              <div class="profile-logo-tweet">
```
Tibor22, line 207: I like how specific you naming classes

index.html
```diff
@@ -59,7 +59,57 @@
             <div class="tweet-button">Button</div>
           </div>
```
AlexDjangoX, line 173: <a href="#" class="section-one-nav__tweet-button">Tweet</a>

Can be styled for a button

Hoolio333: Not sure what you mean here?

AlexDjangoX: I think this a link on twiiter, so it's not a button (which has an action like submitting a form). You can style an a-tag to look like a button

PR #24, opened 2022-02-18
Dan Northcott

Dan2024:


style.css
```diff
@@ -199,7 +202,7 @@ h2 {
 }
 
 .grey-sub-text {
-  color: #d3d8dd;
+  color: rgba(var(--clr-grey));
```
Dan2024, line 205: I'm glad i added a new colour variable here because i used the grey a lot

style.css
```diff
@@ -357,3 +300,57 @@ main .icon-md {
...
+  .nav-list {
+    display: block;
+  }
+
+  .container {
+    display: grid;
+    grid-template-columns: auto 1fr;
+  }
+}
+
```
Dan2024, line 342: was this the best way to add in media queries/responsiveness to website?

style.css
```diff
@@ -357,3 +300,57 @@ main .icon-md {
...
+  .nav-list,
+  .nav-list a {
+    display: block;
+  }
+
+  .container {
+    display: grid;
+    grid-template-columns: 0.3fr 1fr;
+  }
+}
```
Dan2024, line 355: sometimes it's best to not use utility classes when trying to make an area responsive

style.css
```diff
@@ -70,7 +70,7 @@ select {
 /* general */
 
 .center-wrapper {
-  padding-inline: 2em;
+  padding-inline: 1.5em;
```
Dan2024, line 66: nice to use this instead of `padding: 0 1.5em;`

style.css
```diff
@@ -279,3 +301,59 @@ ul {
...
+  padding: 20px;
+}
+
+.story {
+  border-bottom: rgb(170, 169, 169) solid 1px;
+  padding-bottom: 10px;
+  margin-bottom: 10px;
+}
+
+.stories div:last-child {
```
Dan2024, line 308: i liked the use of this selector

style.css
```diff
@@ -0,0 +1,281 @@
...
+  max-width: 100%;
+  display: block;
+}
+
+input,
+button,
+textarea,
+select {
+  font: inherit;
+}
```
Dan2024, line 43: i liked this reset because it meant i didn't have to redefine font families

Tibor22: That inherit property is nice

style.css
```diff
@@ -0,0 +1,376 @@
...
+  padding: 5px;
+  margin-left: 10px;
+  width: 80px;
+  border: none;
+  margin-inline: 10px;
+}
+
+.icon-ex-sm {
+  height: 15px;
+  aspect-ratio: 1;
```
Tibor22, line 140: what is aspect-ratio solve here?

style.css
```diff
@@ -0,0 +1,376 @@
...
+.post-img {
+  width: 90%;
+  border-radius: 15px;
+}
+/* ------------------- */
+/* right styling       */
+/* ------------------- */
+
+.right-section-wrapper {
+  margin-left: 1.5rem;
```
Tibor22, line 282: Why did you choose rem over pixel?

style.css
```diff
@@ -0,0 +1,376 @@
...
+
+/* ------------------- */
+/* Utility classes     */
+/* ------------------- */
+
+/* general */
+
+.center-wrapper {
+  padding-inline: 1.5em;
+  margin-inline: auto;
```
Tibor22, line 67: What these two help you with?

style.css
```diff
@@ -0,0 +1,376 @@
...
+/* ------------------- */
+/* Custom properties   */
+/* ------------------- */
+
+:root {
+  --clr-dark: 0, 22, 50;
+  --clr-grey: 189, 193, 197;
+  --clr-grey-light: 211, 211, 211;
+  --clr-accent: 85, 172, 238;
+  --clr-white: 255, 255, 255;
```
Tibor22, line 54: nice that you using variables

style.css
```diff
@@ -216,13 +228,16 @@ select {
   height: 25px;
   min-width: no25pxne;
   border-radius: 12.5px;
+  margin-right: 10px;
 }
 
 .icon-lg {
   height: 40px;
   aspect-ratio: 1;
```
Hoolio333, line 158: Never heard of aspect ratio - what does that do?

index.html
```diff
@@ -0,0 +1,161 @@
...
+                <ul class="nav-list">
+                    <li><img class="icon-md" src="/assets/generic.svg" alt="generic icon"><a href="#">Home</a></li>
+                    <li><img class="icon-md" src="/assets/generic.svg" alt="generic icon"><a href="#">Explore</a></li>
+                    <li><img class="icon-md" src="/assets/generic.svg" alt="generic icon"><a href="#">Notification</a></li>
+                    <li><img class="icon-md" src="/assets/generic.svg" alt="generic icon"><a href="#">Messages</a></li>
+                    <li><img class="icon-md" src="/assets/generic.svg" alt="generic icon"><a href="#">Bookmarks</a></li>
+                    <li><img class="icon-md" src="/assets/generic.svg" alt="generic icon"><a href="#">Lists</a></li>
+                    <li><img class="icon-md" src="/assets/generic.svg" alt="generic icon"><a href="#">Profile</a></li>
+                    <li><img class="icon-md" src="/assets/generic.svg" alt="generic icon"><a href="#">More</a></li>
+                </ul>
```
Hoolio333: Looks really clean & clear. Nice!

style.css
```diff
@@ -109,89 +92,32 @@ select {
...
-}
-
-.fs-sm {
-  font-size: var(--fs-sm);
-}
-.fs-md {
-  font-size: var(--fs-md);
-}
-.fs-lg {
-  font-size: var(--fs-lg);
```
Hoolio333: I like it how you weren't scared to change or remove so much code. Not needed or right - amend or get rid. No point keeping wrong or pointless code.

style.css
```diff
@@ -0,0 +1,376 @@
...
+  .nav-list a,
+  .right-section {
+    display: block;
+  }
+
+  .container {
+    display: grid;
+    grid-template-columns: 0.4fr 1fr 0.5fr;
+  }
+}
```
PBara7a, line 376: Did you built the page considering mobile first, or is it just easier to visualize your media queries starting from smaller screens?

index.html
```diff
@@ -0,0 +1,159 @@
...
+                    <ul class="nav-list">
+                        <li><img class="icon-md" src="/assets/generic.svg" alt="generic icon"><a href="#">Home</a></li>
+                        <li><img class="icon-md" src="/assets/generic.svg" alt="generic icon"><a href="#">Explore</a></li>
+                        <li><img class="icon-md" src="/assets/generic.svg" alt="generic icon"><a href="#">Notification</a></li>
+                        <li><img class="icon-md" src="/assets/generic.svg" alt="generic icon"><a href="#">Messages</a></li>
+                        <li><img class="icon-md" src="/assets/generic.svg" alt="generic icon"><a href="#">Bookmarks</a></li>
+                        <li><img class="icon-md" src="/assets/generic.svg" alt="generic icon"><a href="#">Lists</a></li>
+                        <li><img class="icon-md" src="/assets/generic.svg" alt="generic icon"><a href="#">Profile</a></li>
+                        <li><img class="icon-md" src="/assets/generic.svg" alt="generic icon"><a href="#">More</a></li>
+                    </ul>
```
PBara7a, line 24: As your descriptive class name suggests, this could be improved by being wrapped inside a `<nav>`.

style.css
```diff
@@ -0,0 +1,376 @@
...
+
+:root {
+  --clr-dark: 0, 22, 50;
+  --clr-grey: 189, 193, 197;
+  --clr-grey-light: 211, 211, 211;
+  --clr-accent: 85, 172, 238;
+  --clr-white: 255, 255, 255;
+
+  --ff-global: "poppins", sans-serif;
+}
```
PBara7a, line 57: Nice! I found the variables very useful as well. I was unsure about where to put them, why the `:root` selector?

index.html
```diff
@@ -4,7 +4,7 @@
     <meta charset="UTF-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <title>Twitter Layout Challenge DN</title>
+    <title>Twitter Layout Challenge - DN</title>
```
Hoolio333, line 7: You don't have many commits.  This makes reviewing harder as lots of changes done in each commit. Also think the commit comments could be more detailed to help know exact changes.

style.css
```diff
@@ -0,0 +1,376 @@
...
+
+.three-row-grid__expand-two {
+  display: grid;
+  grid-template-rows: auto 1fr auto;
+}
+
+.place-items-center {
+  display: grid;
+  place-items: center;
+}
```
PBara7a, line 109: Good layouts and class names, easy to visualize just reading through the code.

style.css
Dan2024: for accessibility reasons - I believe if someone wants the text larger or more zoomed in the rem value will increase inline with the settings they've set but i dont think the same happens when you hard code a pixel value.

style.css
Dan2024: it might be unnecessary, i just wanted to make sure its stayed as a perfect square - i guess i could've just set a width of 15px instead.

style.css
Dan2024: padding-inline margin-inline basically just add what ever value you set to the left and right side.
instead of writing `margin: 0 auto` you can write `margin-inline: auto` i believe it does the same thing. 
you can also have margin/padding/border-block which just adds to the top and bottom.

style.css
Dan2024: my understanding is that it just makes sure the aspect ratio is (in this case)  1/1 ie the icon is a perfect square.  
having an aspect ratio of say 1/1.5 `aspect-ratio: 1/1.5` would mean its 1.5x longer than it is wide i think.

style.css
Dan2024: i didnt start mobile first - it was just easier to visualise starting from mobile first for me. 
In hindsight i think i would rather build the mobile part first though.

style.css
Dan2024: i believe the root selector is selecting the html tag but just with higher specificity. i was under the impression it was standard to use root so thats why i used it haha

PR #36, opened 2022-02-22
Steven Rico

ghost:


index.html
```diff
@@ -0,0 +1,66 @@
...
+  <head>
+    <meta charset="UTF-8" />
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1.0"
+    />
+    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <link rel="stylesheet" href="/styles/reset.css" />
+    <link rel="stylesheet" href="/styles/index.css" />
+    <title>Static Template</title>
```
ghost, line 12: It would be good to add a more appropriate title for this page.

Suggestion: `<title>Twitter</title>`

index.html
```diff
@@ -0,0 +1,66 @@
...
+        <div class="box-md"></div>
+        <nav>
+          <ul class="nav-list">
+            <li>
+              <a class="nav-list--link" href="#">
+                <div class="frame">
+                  <img src="/assets/home.png" alt="home icon" />
+                </div>
+                <div class="link-text">Home</div>
+              </a>
```
ghost: I really like the HTML structure. By wrapping the `img` and the `div.link-text` it means when a user clicks either of them it will re-direct them to the relevant page.

styles/index.css
```diff
@@ -0,0 +1,109 @@
...
+
+  .main-header {
+    width: auto;
+  }
+
+  .nav-list--link {
+    grid-template-columns: auto;
+
+    gap: 0;
+  }
```
ghost, line 168: @Tibor22 love what you said about"re-usable classes sometimes aren't fit for purpose".

This was a good example of having to create a custom class to get the desired effect.

index.html
```diff
@@ -0,0 +1,66 @@
...
+            <div class="button-text">Tweet</div>
+          </button>
+        </div>
+      </header>
+      <main>
+        <!-- DOES IT BREAK THE NATURAL FLOW? WILL THE CONTENT PROVIDE THE HEIGHT? -->
+      </main>
+      <aside class="resources-sidebar">
+        <!-- DOES IT BREAK THE NATURAL FLOW? WILL THE CONTENT PROVIDE THE HEIGHT? -->
+      </aside>
```
ghost, line 134: I don't like this class name. @dogezen or @shahzadnaeem do you have a suggestion?

dogezen: The class represents the container for the whole aside, so something reflecting that would be good:

`.news-aside-container` or similar could be a suggestion

shahzadnaeem: I'd consider `.search-and-suggestions-aside` but also like @dogezen suggestion. Naming can be tricky and trying to be consistent helps.

PR #38, opened 2022-05-04
twitter clone challenge

JohnXander:


gybgivee:
3 things I like
- look just like twitter
- love that each section has their own header.
- reponsiveness
3 things could improve
- maybe leftbar header and rightbar fixed so it looks even more like twitter.
- more cats!
- n/a
3 questions
- why cat ?
- why purple ?
- na

JohnXander:
>   3 questions
> * why cat ? 
Because it's the @boolean-uk/cohort-six mascot 
> * why purple ?
Because I think people like this colour. It's called rebeccaPurple
> * na

miresque:
3 things I like:
1. Nice thought-out layout
2. Clean look with nice button responses
3. I like the structure of the miaow make tweet area

3 things that should be improved:
1. Maybe set a max width so the page doesn't stretch too much
2. Fix Responsiveness by hiding overflowing content
3. Use more Semantic HTML

3 questions:
1. Why use rem & em instead of px?
2. Why cats?
3. Why font awesome?

JohnXander:
> 3 questions:
> 
> 1. Why use rem & em instead of px?
It's more responsive. 
> 2. Why cats?
Because they are the mascot of @boolean-uk/cohort-six 
> 3. Why font awesome?
I noticed a lot of YouTubers using and recommending it.

PR #45, opened 2022-05-06
Twitter clone challenge 

rels742:


Noemi890:
I really like the smoothness of the hovers.
Love the emojis on the main feed.
I like that the buttons change color when you hover. 

I would like to know more about the transition on the hover. 

I think you could do better with the hovers to not cut the icons while your passing over them.
Try do add the border on the main page instead of the right column so it's not going to overflow when your resizing the window.

tuesdah:
Button change when hovering - beautiful
Clone looks fantastic! 

- When typing in the search bar, fix the typing in the middle. 

:) !

PR #52, opened 2022-08-02
Alexander Lind

AlexLind:


Klimcuks (commented):
good job on keeping the code concise and clean.
nice use of sizing columns using (minmax) property.

PR #53, opened 2022-08-04
James Byrne

JamesB94:


Paolo-Federle (commented):
A clawsome work!

TwitterClone/twitter.html
```diff
@@ -0,0 +1,195 @@
...
+      <a href="#"><p>Profile</p></a>
+      
+    </div>
+    <div class="navagation">
+      <i class="fa-solid fa-ellipsis"></i>
+      <a href="#"><p>More</p></a>
+      
+    </div>
+
+    <div class="tweet">
```
Paolo-Federle, line 58: I noticed that when hovering over it from below it enters a strange loop where he tries to scale down and up! But I don't  know why or how to help with that

JamesB94: good spot, don't know why that is will have to check it

PR #54, opened 2022-08-04
Matt Holt

MattVwaves:


index.html
```diff
@@ -0,0 +1,123 @@
...
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <link rel="stylesheet" href="style.css" />
+    <title>Image Grid</title>
+  </head>
+  <body>
+    <div class="container">
+        <div class="left-bar">
```
Paolo-Federle, line 12: I really like how you customized the left bar with the graphics!

style.css
```diff
@@ -0,0 +1,162 @@
...
+}
+ul {
+    list-style: none;
+}
+.container {
+    display: grid;
+    grid-template: 
+        "left-bar main right-bar" 1fr /
+        1fr 700px 1fr;
+    margin-left: 400px;
```
Paolo-Federle: I noticed that I can't see that well the right column because I wasn't using at the full screen the windows of my browser. To help with that (or users with a small screen) I thought about using an auto margin an putting  the container at the  center of the page

mbalabanov:
The overall three column layout is spot-on. Love the names of the Egyptian deities. The whole page seems to be shifted a bit too far to the right.  When I change the view port width, the page does not change (maybe I'm missing something in the responsive behaviour).

PR #55, opened 2022-08-04
Patrik Ellini

PatrikE91:


dearshrewdwit (commented):
Great choice of colour for your elements!

style.css
```diff
@@ -0,0 +1,267 @@
...
+    margin-top: 4px;
+}
+.header div, .header .lence{
+    align-self: center;
+}
+.search{
+    display: grid;
+    grid-template-columns: 30px 1fr;
+    background-color: #eceef0;
+    margin-left: 10px;
```
Klimcuks, line 57: here, to keep you code concise, you could use a shorthand property to set padding and margin

Klimcuks (commented):
Well done on making the code uncluttered and easy to read. To improve that even more, use html structure elements like heading, article etc.
Good use of class selectors.

Paolo-Federle (commented):
Love the vibes.
Very good job, having the middle (and most important) column that resize with the page is a great idea to avoid the risk of having to much margin on the sides. 
Great use of the code even for placing the icons for the posts without having a grid, I couldn't do it. 
100 punti su 100.

PR #57, opened 2022-08-05
pooja

dhanwadepooja2688:


twitter.css
```diff
@@ -0,0 +1,110 @@
...
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+  font-family: Arial, Helvetica, sans-serif;
+}
+
+.twitter {
+  display: grid;
+  grid-template:
+    "left main search" /
```
AlexLind, line 11: You should consider having some whitespace on the left/right side of the page.

twitter.css
```diff
@@ -0,0 +1,110 @@
...
+    1fr 2fr 1fr;
```
AlexLind, line 12: Consider fixing the width of the the left and search area and only using fractions for the main part.

twitter.css
```diff
@@ -0,0 +1,110 @@
...
+
+.image {
+  grid-area: image;
+  text-align: center;
+}
+
+.image img {
+  border-radius: 50%;
+}
+
```
AlexLind, line 107: Consider adding some media querys to make the site responsive

AlexLind (commented):
Good job Pooja! It almost looks like the real twitter.

mbalabanov:
Looks good. The column layout matches Twitter's. Uses columns within columns. In the main sections, everything is where it should be. Some of the colors and icons might benefit from a second look and the right column looks like it could also use some attention. Maybe I'm missing something but the page doesn't seem to be responsive.

PR #60, opened 2022-08-22
Twitter-Paolo Federle

Paolo-Federle:


mbalabanov:
Awesome solution with a great use of columns within columns. The colors and the look and feel are spot on and look like the original Twitter. Some responsive changes happening with a different background color at a narrower view port. Overall a really good job!

PR #70, opened 2022-10-07
Twitter Challange

akibria-dev:


style.css
```diff
@@ -0,0 +1,150 @@
...
+    margin: 0;
+    box-sizing: border-box;
+}
+
+div {
+    font-size: 10px;
+    font-family: 'Segoe UI', Tahoma, Geneva, Verdana, sans-serif;
+}
+
+.container {
```
Joelzor: Obviously not much to comment on as it's not finished but the main layout looks good, looking forward to seeing the rest of it :)

PR #71, opened 2022-10-07
twitter challenge

mlync87:
Been watching a lot of youtube on how to do this task. this process doesn't involve grids, but i found it very helpful in learning how to allocate buttons in the sidebar, as well as coloring text and creating active links. After some more research on grids, i think i can now build the grid needed and incorporate these elements into it.


mlync87:
I've added the header, image and button. I still haven't used grid format and my header is slightly to the left. once I complete the second side bar and where the posts go, I'm planning on putting the grid in using all the declared items and deleting some of the details from the html to make it simpler. Getting the hang of putting buttons and using margin: auto to move it the desired area.

mlync87:
just added a post with an image, verified tag, and icons along bottom. Getting a much better idea of how to deal with individual elements and how to position them.

mlync87:
ok, I've all the elements in place. I just need to work make a few edits. Then I can construct my grid and place all elements inside it.   By constructing this page without a grid i have learned a lot about declaring individual elements and how to use stylesheet to edit their appearance etc

mlync87:
added the last few icons. just need to work on spacing them properly, and then i can go back and reconfigure everything into grid format

mlync87:
Inserted a grid into css. still need to go through some elements on html and rename them to make them easier to work with. Having done this assignment without using css grid initially, and then working backwards inserting a grid and changing settings, i find i am retaining information better. Just annoying that it took me 18 hours to figure that out.

twitter.html
```diff
@@ -0,0 +1,188 @@
...
+                    Martin Lynch
+                    <span class="post-headerSpecial">
+                      <!--attach the verified symbol via link-->
+                      <span class="material-symbols-outlined"> verified </span> @MartinLynch
+                    </span>                    
+                  </h3>
+              </div>
+              <!-- assign text that will appear as a post-->
+              <div class ="post-headerDescription">
+                <p>Good lord am i finding this assignment difficult lol</p>
```
lucianosimoni, line 119: 😂😂😂

PR #83, opened 2022-12-09
Manoj-Twitter challenge

Manojkumar883:
almost completed, need to change some adjustments

### DIFF
--- a/style.css
+++ b/style.css
@@ -1,0 +1,342 @@
+*{
+    font-family:Arial, Helvetica, sans-serif;
+    margin:10px;
+    background-color: rgb(39, 39, 50);
+    color:white;
+    scroll-behavior: smooth;
+
+}
+/* main container */
+.container {
+    display:grid;
+    grid-template:
+    "left-menu main right-menu" 1fr /
+    0.35fr        1fr       0.5fr;
+    grid-gap:5px;
+    margin:0
+}
+/* left-menu */
+.left-menu{
+    grid-area:left-menu;
+    display:grid;
+    margin:10px;
+    grid-gap:10px
+
+}
+/* left-menu-icons */
+.left-menu .icons {
+    display:grid;
+    grid-template:
+    "Home"0.1fr
+    "Explore" 0.1fr
+    "Notifications"0.1fr
+    "Messages"0.1fr
+    "Bookmarks"0.1fr
+    "lists"0.1fr
+    "profile"0.1fr
+    "More"0.3fr/
+    1fr;
+}
+/* logo image*/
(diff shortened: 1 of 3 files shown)


twitter.html
```diff
@@ -0,0 +1,139 @@
...
+    <script src="https://kit.fontawesome.com/59ab1ef207.js" crossorigin="anonymous"></script>
+    <link rel = "stylesheet" href ="style.css" />
+</head>
+<body>
+    <div class="container">
+        <div class="left-menu"> 
+            <div class="logo">
+                <img src = "https://upload.wikimedia.org/wikipedia/commons/thumb/4/4f/Twitter-logo.svg/512px-Twitter-logo.svg.png?20220821125553"  alt = "logo"/>
+            </div>
+            <div class="icons">
```
WestyCodes: Your icons, and the hover use is fantastic... are the icons just a font family? I have no idea how this works!

twitter.html
```diff
@@ -0,0 +1,139 @@
...
+                <div class="main-Home">
+                    <p><strong>Home</strong></p>
+                    <i class="fa-regular fa-star"></i>
+                </div>
+                <div class="profile-image">
+                    <img src = "https://cn.i.cdn.ti-platform.com/cnapac/content/2017/showpage/ben-10/sa/showicon.png" alt = "profileImage" />
+                    <span>What's happening?</span>
+                </div>
+                <div class="upload">
+                    <div class="H-icons">
```
WestyCodes, line 61: Here is that font-awesome again! It is blowing my mind.

style.css
```diff
@@ -0,0 +1,342 @@
...
+    margin:0;
+    font-size: 15px;
+    border-radius: 15px;
+    background-color: #1D9BF0;
+    color: white;
+    font-weight: 700;
+    border: 1px solid #1D9BF0;
+
+}
+
```
WestyCodes, line 79: The colour choice and scale for your hover is super aesthetically pleasing!

twitter.html
WestyCodes: The only possible improvement I would suggest, is putting the anchor tags in a nav tag for accessibility. Either way - fantastic job bud!

WestyCodes (commented):
10/10 would read this code again!

twitter.html
Manojkumar883: Hey, bud. there is a website called "fontawesome" . if we import the link of the page in to our html. we can use some the free icons directly in our code.

Manojkumar883: That's a really good point.  i totally forgot to differentiate the sections in html.

PR #91, opened 2023-03-29
The Alistair Henshaw

Alistair1080:


twitter.html
```diff
@@ -0,0 +1,150 @@
...
+    </head>
+    <body>
+        <div class="layout">
+            <aside class="left">
+                <img src="./icons/twitter-app-icon.svg" alt="twitter logo" class="logo">
+                <nav class="nav">
+                    <a href="#main" class="hom bar" tabindex="1">Home</a>
+                    <a href="explore.html" class="exp bar" tabindex="2">Explore</a>
+                    <a href="notifications.html" class="not bar" tabindex="3">Notifications</a>
+                    <a href="messages.html" class="mes bar" tabindex="4">Messages</a>
```
timmyctt, line 15: Have not seen tableindex before. What does it do?

carolarruda: Yes, spot on Timmy. What does that do?

profile.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>Profile</title>
+        <link rel="stylesheet" href="style.css">
+    </head>
+    <body>
+        <h1>
+            Looking fresh. The world needs more of your mirror selfies.
```
carolarruda, line 9: These were funny. Had fun exploring the page and finding the gems

twitter.html
Alistair1080: elements which aren't normally able to be selected using the tab bar, can be given this so when it gains the ability to be navigated to on the browser using tab.

profile.html
Alistair1080: haha thanks, helped me enjoy the challenge more

PR #92, opened 2023-03-29
completed core + extension task (korey)

rockettdev:


index.html
```diff
@@ -0,0 +1,145 @@
...
+    <div class="grid-container">
+        <div class="left-sidebar">
+            <div class="leftbar twitterlogo"> <img src="/images/svg/twitter.svg"></div>
+            <div class="leftbar lefticon"><img src="/images/svg/home.svg"><p>Home</p></div>
+            <div class="leftbar lefticon"><img src="/images/svg/explore.svg"><p>Explore</p></div>
+            <div class="leftbar lefticon"><img src="/images/svg/notifications.svg"><p>Notifications</p></div>
+            <div class="leftbar lefticon"><img src="/images/svg/messages.svg"><p>Messages</p></div>
+            <div class="leftbar lefticon"><img src="/images/svg/bookmarks.svg"><p>Bookmarks</p></div>
+            <div class="leftbar lefticon"><img src="/images/svg/profile.svg"><p>Profile</p></div>
+            <div class="leftbar lefticon"><img src="/images/svg/more.svg"><p>More</p></div>
```
NoroAxper, line 21: Perhaps instead of Divs use Buttons, this is for correct semantics also you would be able to focus on the buttons using the Tab key whereas divs don't get focused on

index.html
```diff
@@ -0,0 +1,145 @@
...
+                <div class="whatshappening">What's Happening?</div>
+                <div class="icontweet">
+                    <div class="media"><img src="/images/svg/media.svg"></div>
+                    <div class="gif"><img src="/images/svg/gif.svg"></div>
+                    <div class="stats"><img src="/images/svg/stats.svg"></div>
+                    <div class="emotes"><img src="/images/svg/emotes.svg"></div>
+                    <div class="calendar"><img src="/images/svg/calendar.svg"></div>
+                    <button>Tweet</button>
+                </div>
+
```
NoroAxper, line 39: I like that you spaced things out, even this 1 line makes it so much cleaner and easier to read, i definitely need to implement something like this, good luck to my reviewers haha

index.html
```diff
@@ -0,0 +1,145 @@
...
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Home / Twitter</title>
+    <link rel="stylesheet" href="style.css">
+</head>
+<body>
+    
+    <div class="grid-container">
+        <div class="left-sidebar">
```
NoroAxper, line 13: instead of Div you could use Nav tag, again for semantics

style.css
```diff
@@ -0,0 +1,284 @@
...
+    grid-template-rows: 75px 140px 20px repeat(6, 130px);
+    border-bottom: solid 1px  #dbdbdb;
+}
+
+.right-sidebar {
+    grid-area: right;
+    background-color: rgb(255, 255, 255);
+    text-align: center;
+    display: grid;
+    grid-template-columns: 1fr;
```
NoroAxper, line 37: There are properties which are repeated, instead you could re use classes to write less lines,
For example if you did 
.left-sidebar, .main, .right-sidebar {
background color;
display grid;
grid tempplate columns 1fr;
}

style.css
```diff
@@ -0,0 +1,284 @@
...
+    margin-left: 10px;
+}
+
+.search > p {
+    padding: 8px;
+    font-weight: 500;
+    text-align: left;
+}
+
+
```
NoroAxper, line 259: Great work bro, very clean and easy to read!

PR #97, opened 2023-03-30
Todd Griffin

ample-samples:


index.html
```diff
@@ -0,0 +1,112 @@
...
+      <div id="explore"><img src="assets/icons/hashtag-icon.svg" alt="">Explore</div>
+      <div id="notifications"><img src="assets/icons/notification-bell-icon.svg" alt="">Notifications</div>
+      <div id="messages"><img src="assets/icons/message-icon.svg" alt="">Messages</div>
+      <div id="bookmarks"><img src="assets/icons/bookmark-bar-icon.svg" alt="">Bookmarks</div>
+      <div id="lists"><img src="assets/icons/note-icon.svg" alt="">Lists</div>
+      <div id="profile"><img src="assets/icons/user-profile-icon.svg" alt="">Profile</div>
+      <div id="more"><img src="assets/icons/more-options-ellipsis-icon.svg" alt="">More</div>
+      <div class="tweet-button">Tweet</div>
+      <div></div>
+      <div></div>
```
marc-alzapiedi, line 23: It seems like these empty divs are being used to make some kind of space. Interesting use of the divs

index.html
```diff
@@ -0,0 +1,112 @@
...
+      </div>
+      <div id="home"><img src="assets/icons/homepage-icon.svg" alt="">Home</div>
+      <div id="explore"><img src="assets/icons/hashtag-icon.svg" alt="">Explore</div>
+      <div id="notifications"><img src="assets/icons/notification-bell-icon.svg" alt="">Notifications</div>
+      <div id="messages"><img src="assets/icons/message-icon.svg" alt="">Messages</div>
+      <div id="bookmarks"><img src="assets/icons/bookmark-bar-icon.svg" alt="">Bookmarks</div>
+      <div id="lists"><img src="assets/icons/note-icon.svg" alt="">Lists</div>
```
marc-alzapiedi, line 18: It seems like you have a bunch of svg's that you have saved on your local machine and then you have used the img tag to include them in the page instead of having really long svg tags. I didn't think of doing this it would of probably made my code much shorter.

index.html
marc-alzapiedi: Or maybe the empty divs are being used to construct grids, please excuse me if i'm wrong

index.html
```diff
@@ -0,0 +1,112 @@
...
+        <hr>
+      </div>
+
+    </section>
+
+    <section class="happening"> 
+      <div>Search</div>
+      <div>What's happening<hr></div>
+      <div>Who to follow<hr></div>
+
```
marc-alzapiedi, line 125: There are a lot of empty divs's

style.css
```diff
@@ -0,0 +1,206 @@
+* {
+  margin: 0;
+  padding: 0;
+  font-weight: bold;
```
marc-alzapiedi: Doesn't this make all the contents in the page bold? Interesting start to style the page

style.css
```diff
@@ -0,0 +1,206 @@
...
+
+.main .compose-grid > .compose-post {
+  display: grid;
+  grid-template-rows: 1fr 1fr;
+  border: none;
+}
+
+.main .compose-grid > .compose-post > .post-attachments {
+  display: grid;
+  grid-template-columns: 40px 40px 40px 40px 40px 1fr 200px;
```
marc-alzapiedi, line 174: Here I would say is a good example to take advantage of the use of repeat() where there are the same dimensions for multiple grid elements, maybe.....

style.css
```diff
@@ -0,0 +1,206 @@
...
+.happening {
+  display: grid;
+  grid-template-rows: repeat(auto-fit, 60px);
+}
+
+.happening > div {
+  margin: 4px;
+  border: 3px solid black;
+  border-radius: 7px;
+}
```
marc-alzapiedi, line 254: I think it looks nice overall but the code of the html and CSS is strikingly short. At the end I ask myself if the twitter page is complete? Or is there still more to add/adjust/finish?

The length of the code makes it hard to believe that you have finished, and if you have finished everything then hats off to you very well done!

ample-samples: Hi Marc, thought the structure is all here, there's definitely a bit more to add content-wise.

index.html
ample-samples: True, I suppose I could've used p tags here instead

index.html
alexjshaw: Having looked at your code Marc, I can say for sure that yes it would have!

index.html
```diff
@@ -0,0 +1,112 @@
...
+      </div>
+      <div id="home"><img src="assets/icons/homepage-icon.svg" alt="">Home</div>
+      <div id="explore"><img src="assets/icons/hashtag-icon.svg" alt="">Explore</div>
+      <div id="notifications"><img src="assets/icons/notification-bell-icon.svg" alt="">Notifications</div>
+      <div id="messages"><img src="assets/icons/message-icon.svg" alt="">Messages</div>
+      <div id="bookmarks"><img src="assets/icons/bookmark-bar-icon.svg" alt="">Bookmarks</div>
+      <div id="lists"><img src="assets/icons/note-icon.svg" alt="">Lists</div>
+      <div id="profile"><img src="assets/icons/user-profile-icon.svg" alt="">Profile</div>
+      <div id="more"><img src="assets/icons/more-options-ellipsis-icon.svg" alt="">More</div>
+      <div class="tweet-button">Tweet</div>
```
alexjshaw, line 21: I would probably suggest that you use a p tag or span tag for your text here. Good practice to be able to target the text directly if you want to change the font size, color, boldness, or anything else.

style.css
```diff
@@ -0,0 +1,206 @@
...
+
+.navigation > div + div + div + div + div + div + div + div + div + div + div {
+  z-index: -1;
+}
+.navigation > div + div + div + div + div + div + div + div + div + div + div:hover {
+  background-color: white;
+}
+.navigation > div + div + div + div + div + div + div + div + div + div + div:active {
+  background-color: white;
+}
```
alexjshaw, line 83: I feel like you know this is bad! I get what you're doing (after you explained it) but even so you could have used nth:child to make it much cleaner.

index.html
```diff
@@ -0,0 +1,112 @@
...
+      <div></div>
+      <div></div>
+    </section>
+
+    <section class="main">
+      <div>Home</div>
+      <div class="compose-grid">
+        <div class="profile-pic"></div>
+        <div class="compose-post">
+          <div class="enter-text" contenteditable>What's new?</div>
```
alexjshaw: using input here would be more appropriate rather than an editable div.

style.css
Kocibey01: Nice work mate, very clean and easy to read! 💯

index.html
```diff
@@ -0,0 +1,112 @@
...
+          <div class="post-attachments">
+            <div><img src="assets/icons/picture-icon.svg" alt=""></div>
+            <div><img src="assets/icons/gif-file-icon.svg" alt=""></div>
+            <div><img src="assets/icons/poll-icon.svg" alt=""></div>
+            <div><img src="assets/icons/happy-icon.svg" alt=""></div>
+            <div><img src="assets/icons/information-date-calendar-icon.svg" alt=""></div>
+            <div></div>
```
alexjshaw, line 38: Why have you got all of your images wrapped in blank divs? While you may not have needed to modify them in your CSS beyond the grid-layout for post-attachments it would be good practice to at least have them in a div with a class of "icon-container" or something like that so you can easily make changes later on.

alexjshaw (commented):
Good effort Todd, but please use comments more!

Just think about if you came back to this in a couple of months time. How easy would it be to read and understand what each section is? A few comments, just to mark the start of each section, would go a long way to making it more understandable.

Kocibey01 (commented):
Nice work mate, very clean and easy to read!

PR #102, opened 2023-03-31
Twitter Challenge - Priya

ps975076:


styles.css
```diff
@@ -0,0 +1,397 @@
...
+  margin-left: 143px;
+}
+
+.media-upload:hover {
+  cursor: pointer;
+}
+
+.gif-upload {
+  width: 25px;
+  position: relative;
```
ps975076, line 360: Used relative positioning for all media icons. Is there a better way of positioning icon?

index.html
```diff
@@ -0,0 +1,117 @@
...
+
+      <!--Right Side-Bar Starts-->
+      <div class="right-bar">
+        <div class="magnifying">
+              <img src="./magnifier-glass-icon (1).svg">
+            </div> 
+          <div class="search-twitter-bar">Search Twitter
+             
+          </div>
+          <div class="search-whats-happening"><h2>What's Happening</h2><p><span>Trending in United Kingdom     </span></p><p>Usyk</p><p>32.8K Tweets</p><p><span>Politics &#183 Trending</span></p><p>Council Tax</p><p>7,400 Tweets</p><p><span>News &#183 Trending</span></p><p>Owen Jones</p><p>2,054 Tweets</p></div>
```
ps975076: Is there a way to close gaps between paragraphs?

index.html
```diff
@@ -0,0 +1,117 @@
...
+            </div>
+          </div>
+          <!--Add div for button-->
+            <button class="tweet-button2">Tweet</button>
+       </div>
+       <hr>
+       <!--Main - last section feed1-->
+       <div class="feed">
+        <div class="feed-grid1-profile-pic">
+          <img src="./default-profile-picture-grey-female-icon.svg">
```
ps975076, line 116: Inserted an icon SVG instead of an img, when I did insert an img - I was unable to downsize, I believe the type of img was a png, is this a correct file type?

index.html
```diff
@@ -0,0 +1,117 @@
...
+    <link rel="stylesheet" href="styles.css" />
+  </head>
+  <body>
+    <!--Grid starts-->
+    <div class="grid">
+      <!--Left Side-Bar Starts-->
+      <div class="left-bar">
+        <!--Twitter Logo on left-bar-->
+        <div class="logo-grid-leftbar">
+
```
alexjshaw: I like how clearly you comment your code, makes it nice and easy to follow and is a good practice as things start to scale up.

styles.css
```diff
@@ -0,0 +1,397 @@
...
+
+.explore {
+  grid-area: explore;
+  text-align: left;
+  margin-left: 70px;
+}
+
+.explore:hover {
+  cursor: pointer;
+}
```
alexjshaw, line 111: At the moment this changes your cursor if you hover over anywhre on the entire left column. If you added a second class, such as class: "nav-button" to your individual buttons it would enable you to then set the cursor to only change when you mouse over the buttons themselves.

index.html
```diff
@@ -0,0 +1,117 @@
...
+        <div class="options-nav">
+          <div class="home"><h2>Home</h2></div>
+          <div class="explore"><h2>Explore</h2></div>
+          <div class="notifications"><h2>Notifications</h2></div>
+          <div class="messages"><h2>Messages</h2></div>
+          <div class="bookmarks"><h2>Bookmarks</h2></div>
+          <div class="lists"><h2>Lists</h2></div>
+          <div class="profile"><h2>Profile</h2></div>
+          <div class="more"><h2>More</h2></div>
+        </div>
```
alexjshaw, line 74: Could maybe space this out a bit more.

```
<div class="home">
   <h2>
      Home
   </h2>
</div>
```

index.html
alexjshaw: line-height will allow you to change the spacing between lines

alexjshaw (commented):
Good effort overall Priya.

index.html
ps975076: Cheers Alex - I'll check that out

styles.css
ps975076: Thank you - Will add a class so the cursor only changes when I move to the individual buttons.

PR #103, opened 2023-03-31
Sara twitter challenge

sarajbeazley:


images/twitter.html
```diff
@@ -0,0 +1,72 @@
...
+                <div class="icon">
+                    <img src="https://th.bing.com/th/id/OIP.V-c35hiAW715PW5zj8vVBQHaHa?w=163&h=180&c=7&r=0&o=5&dpr=2&pid=1.7" alt="icon">
+                    <div class="hashtag"><strong>Explore</strong></div>
+                    </div>
+                <div class="icon"><strong>Notifications</strong></div>
+                <div class="icon"><strong>Messages</strong></div>
+                <div class="icon"><strong>Bookmarks</strong></div>
+                <div class="icon"><strong>Lists</strong></div>
+                <div class="icon"><strong>Profile</strong></div>
+                <div class="icon"><strong>More</strong></div>
```
NoroAxper: Maybe use buttons instead of divs for correct semantics so the users can switch between them with the tab key

images/twitter.html
```diff
@@ -0,0 +1,72 @@
...
+
+            </div>
+            <div class="tweet-button">
+                <a href="https://twitter.com/" class="social-link Twitter">Tweet</a>
+
+            </div>
+            <div class="name">@SaraBeazley</div>
+        </div>
+        <!-- Main feed -->
```
NoroAxper, line 62: great use of commenting, very easy to read your work, mine is such a mess haha...

images/style.css
```diff
@@ -0,0 +1,212 @@
...
+    position: relative;
+    left: 60px;
+    top: 20px;
+    /* border: 3px dotted purple; */
+    font-size: larger;
+    color: grey;
+}
+
+.feed {
+    display: grid;
```
NoroAxper, line 167: display: grid; is repeated a lot, you could have a section where all of the classes that are grid are in 1 list and if you add elements which you want to be a grid, just to add it to that list.
or you could do what i mentioned in class just have 1 class for example .grid { display: grid;} then add that grid class in your html to which ever elements you want to be a grid

images/style.css
```diff
@@ -0,0 +1,212 @@
...
+    grid-area: whats-happening;
+    text-align: center;
+    border: 3px dotted purple;
+}
+
+.who-to-follow {
+    grid-area: who-to-follow;
+    text-align: center;
+    border: 3px dotted purple;
+}
```
NoroAxper, line 235: Great work! very clean and easy to read

PR #115, opened 2023-09-27
Kye

yee0802:


images/assets/cat.jpg
Alistair1080, line 1: Excellent choice of picture!

index.html
```diff
@@ -0,0 +1,423 @@
...
+                        </div>
+                    </div>
+
+                    <div class="home-feed-tweet-icon">
+                        <i class="fa-regular fa-comment"></i>
+                        <i class="fa-solid fa-retweet"></i>
+                        <i class="fa-regular fa-heart"></i>
+                        <i class="fa-solid fa-upload"></i>
+                        <i class="fa-regular fa-gem"></i>
+                    </div>
```
Alistair1080, line 89: Nice use of class selectors to apply the right styles which are repetitive!

index.html
```diff
@@ -0,0 +1,423 @@
...
+            </div>
+            <div class="nav-index">
+                <div class="nav-index-inner active"><i class="fa-solid fa-house"></i><h2>Home</h2></div>
+                <div class="nav-index-inner"><i class="fa-solid fa-hashtag"></i><h2>Explore</h2></div>
+                <div class="nav-index-inner"><i class="fa-regular fa-bell"></i><h2>Notifications</h2></div>
+                <div class="nav-index-inner"><i class="fa-regular fa-envelope"></i><h2>Messages</h2></div>
+                <div class="nav-index-inner"><i class="fa-regular fa-rectangle-list"></i><h2>Bookmarks</h2></div>
+                <div class="nav-index-inner"><i class="fa-regular fa-user"></i><h2>Profile</h2></div>
+                <div class="nav-index-inner"><i class="fa-solid fa-ellipsis"></i><h2>More</h2></div>
+            </div>
```
Alistair1080, line 24: I feel this section needs more indentation adding to improve readability

style.css
```diff
@@ -0,0 +1,575 @@
...
+    min-width: 492px;
+}
+
+/* Tweet Bar */
+
+.home-main {
+    display: grid;
+    grid-template: 55px 140px / 1fr;
+    border-bottom: 14px solid #E1E8ED60; 
+}
```
Alistair1080, line 154: I admire your ability to use a variety of units e.g fr, px, rem. However I do wonder if doing this has caused you issues with calculating values?

Alistair1080 (commented):
Nice work fella! A lot to read hear so I can see you've put some time in creating it which is cool. Also good use of media quires to make the site dynamic.

jdm1991 (commented):
Easy to read HTML. Really nice. I also found your CSS really concise and easy to follow. So much better than having it all over the place.

index.html
```diff
@@ -0,0 +1,423 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Twitter</title>
+    <link rel="stylesheet" href="style.css">
```
NazarTymiv, line 7: I think it would be better to contain all of the css files inside css folder :)

index.html
```diff
@@ -0,0 +1,423 @@
...
+    <script src="https://kit.fontawesome.com/367b368e67.js" crossorigin="anonymous"></script>
```
NazarTymiv, line 8: Very good to use icon library, I like it :)

index.html
```diff
@@ -0,0 +1,423 @@
...
+                    </div>
+                    <div class="show-more">
+                        <a href="#">Show More</a>
+                    </div>
+                </div>
+            </div>
+        </div>
+    </main>
+</body>
+</html>
```
NazarTymiv, line 423: You can improve you class names using BEM metodology, for example home is parend element. home__main is children element of home and home__main-item is children element of home__main.

NazarTymiv (commented):
Great work!

style.css
yee0802: Not particularly, 1rem is refrencing the root font-size (which by default is 16px). I have used it for quite a while so that's why  I used it! (Might not be good practice idk)

index.html
yee0802: That's definitely something I will have a look at, I thought my class names were a bit messy overall. Thanks for the insight! (p.s the section you highlighted was rushed last minute haha)

index.html
yee0802: Thanks !

yee0802:
Won't use FontAwesome in future.

PR #117, opened 2023-09-27
Chloe

Chloe070196:


src/style.css
```diff
@@ -0,0 +1,363 @@
...
+    font-size: 10px;
+}
+
+::-webkit-scrollbar {
+    border: none;
+}
+
+/*STYLES THE LEFT-MENU*/
+
+#clone-icon {
```
TomEastwood, line 111: I like how youve switched to use class ID selectors in your left-nav menu

src/style.css
```diff
@@ -0,0 +1,363 @@
...
+/*FOR BIGGER DEVICES (laptops, tablets,...)*/
+
+    #bottom-menu {
+        width: 0;
+        content-visibility: hidden;
+    }
+
+/*FOR SMALLER RESOLUTION DEVICES*/
+
+@media screen and (max-width:480px) {
```
TomEastwood, line 410: This is a great way to get both side-bars to dissapear, Id split mine into two different media queries but this works even better

assets/svg/stormtrooper-svgrepo-com.svg
EwbiDev, line 1: I like the theming! 😄

src/style.css
```diff
@@ -0,0 +1,363 @@
...
+    #left-menu, 
+    #right-menu {
+        padding: 0;
+        margin: 0;
+        width: 0;
+        content-visibility: hidden;
+    }
```
EwbiDev, line 417: Agree with Tom, mobile view looking good.

I'd suggest maybe limiting the amount of icons in the bottom bar and giving them each more space. It could be quite tricky to select them on a small screen.

src/style.css
```diff
@@ -0,0 +1,363 @@
...
+}
+
+footer {
+    font-size: 10px;
+    height: fit-content;
+}
+
+
+
+/*RECURRENT STYLES*/
```
EwbiDev, line 31: Using classes for general attributes, big like 👍

src/style.css
```diff
@@ -0,0 +1,363 @@
...
+}
+
+/*hover and selected styles for LEFT-MENU*/
+#left-menu main > div:hover {
+    color: orange;
+}
+#left-menu main > div:active {
+    background-color: rgba(255, 166, 0, 0.6);
+    color: white;
+}
```
EwbiDev: I like the hover stuff!

The active selector is quite quick though and quite easily missed. I think it would be worth to take a look at whether you can use [transitions](https://developer.mozilla.org/en-US/docs/Web/CSS/transition) to make them linger longer.

src/index.html
```diff
@@ -0,0 +1,301 @@
...
+</body>
+</html>
+
+
+
+<!--☑ Layout the main page elements (left-hand menu, tweet feed and right side-bar) using CSS Grid. CSS Grid should also be used to layout individual tweets.
+☑ Define the page contents using HTML with appropriate elements. HTML should be well formatted.
+☑ Style the page using a mixture of Type, ID and Class CSS selector types
+☑ Implement hover and selected state for the left-hand menu links.-->
```
EwbiDev, line 341: Oooh very organised, my todos were just hidden around my code 😆

EwbiDev (commented):
Very good!

src/style.css
Chloe070196: thanks for that :) nice to get some new stuff to use, I'll definitely look into it!

src/style.css
Chloe070196: ooops, didn't think of that last bit - makes sense though! I'll keep 3, have 'more' as the 4th, with the assumption that the ones that aren't visible can be accessed  if 'more' is open. 

Could I actually do that last bit (having a 'more' menu be accessible once activated) without needing JS?

assets/svg/stormtrooper-svgrepo-com.svg
Chloe070196: I was told to make a 'clone', just couldn't resist haha

src/style.css
Chloe070196: thanks!

src/style.css
Chloe070196: cheers! trying to make my css less confusing than it used to be - may or may not have managed to confuse myself rather thoroughly earlier this week ^^

src/style.css
Chloe070196: thanks!

src/index.html
Chloe070196: honestly, that's just me copy-pasting the todos from the README files so I can glance at them without have to open said file - I'm trying to limit the number of tabs and windows I've got open at any given time, otherwise I lose track of some, then waste time trying to find info. Could blame that on my having only one monitor, but it's probably also got a lot to do with learning to organise stuff lol

src/style.css
EwbiDev: It's kinda possible in just CSS, for my Discord assignment I used [this thread](https://stackoverflow.com/questions/44832572/pure-css-clickable-dropdown) to make dropdowns.

Chloe070196: I'll give that a go then, cheers!

Chloe070196: Found a way, might be kinda similar (mine behaves a bit like a pop up thingy though, and uses :focus (I'm actually not sure how hover behaves with phones, so went for avoiding it)) - not the most elegant, but it seems to work haha

PR #118, opened 2023-09-28
Elizabeth

Elizabethcodes44:


twitter.css
```diff
@@ -0,0 +1,141 @@
+*{  box-sizing: border-box;
+    font-family: sans-serif;
+    
+
+}
+.Container{
+    display: grid;
+    grid-template-columns: 275px auto 350px; /*it can also be written like this instead of writing them out if it has just one column for example no footer and header and other blocks*/
```
EwbiDev, line 8: 100% agree with your comment here! If your grid only has cells all equal size, defining areas isn't strictly necessary.

twitter.css
```diff
@@ -0,0 +1,141 @@
...
+    width:30px;
+}
+.ticon-home img {
+    width: 40px;
+}
+.ticon-home {
+    display: grid;
+    grid-auto-flow: column;
+
+}
```
EwbiDev, line 41: Using a grid layout for the nav icons and the text is a sensible idea here. I would look at expanding it out a bit. You set a certain amount of columns for `.container` so we can attempt a similar thing here like below:

|logo|text|
| - | - |
|logo|text|
|logo|text|
|logo|text|

I would suggest doing something `grid-template-columns: 40px 200px;` and changing the grid flowing back to rows like so: `grid-auto-flow: row;` and seeing what happens when you have more than two elements in your `ticon-home` div.

twitter.css
```diff
@@ -0,0 +1,141 @@
...
+
+.right-menu {
+    outline: 3px solid pink;
+}
+.right {
+   position: fixed;
+   Margin: 10px 10px 10px 10px;
+   display: grid; 
+}
+.Whats-happening {
```
EwbiDev, line 107: Class names are clear and descriptive! I'd be careful to make sure they're all in the same casing and style to make sure you're referencing them properly in the future. In this case the word ought not to be capitalized. Remember to change what's in your HTML as well if you change stuff here!

twitter.html
```diff
@@ -0,0 +1,67 @@
...
+  </head>
+  <body>
+    <div class="Container">
+        <div class="left-menu">
+            <div class="ticon-logo">
+                <img src="./images/5296514_bird_tweet_twitter_twitter logo_icon.png"> 
+            </div>
+            <div class="icons">
+                <div class="ticon-home">
+                    <img src="./images/2246842_home_home page_house_icon.png">
```
EwbiDev, line 18: This solution to your svg woes was along the line of what I would have suggested.

By default, the `background` property in CSS [is set to repeat](https://developer.mozilla.org/en-US/docs/Web/CSS/background#repeat-style) which is why you were getting multiple of them.

twitter.css
```diff
@@ -0,0 +1,141 @@
...
+    outline: 3px solid pink;
+    margin-top: 10px;
+}
+.numbers {
+    display: grid;
+    grid-auto-flow: row;
+    height: 400px;
+    gap: 10px; 
+    margin-top: 5px;
+}
```
EwbiDev, line 120: Separating your grids into various sub-grids is a good way to go! I reckon you can do the same with your tweets as well.

twitter.css
```diff
@@ -0,0 +1,141 @@
...
+.tweet3 {
+    outline: 3px solid pink;
+}
+.tweet3 img {
+    height:30px;
+    width: 30px;
+    border-radius: 50%;
+    
+}
```
EwbiDev, line 173: Instead of having different classes for each of these tweet posts, I would go with all 3 of these having the same class because they will all want the same styling. Maybe `.tweet-post` or something like that?

twitter.html
EwbiDev, line 1: standard convention is your homepage for a site is called `index.html`, exactly why, I don't know that's beyoind my scope of knowledge 😅

EwbiDev (commented):
Keep at it, you'll be flying through this in no time 😇

PR #124, opened 2023-09-28
Ryan

Radio58:


homepage.html
```diff
@@ -0,0 +1,144 @@
...
+              <img src="src/images/frame.svg" width="20px" alt="frame">
+              <img src="src/images/gif.svg" width="20px" alt="gif icon">
+              <img src="src/images/statistics-svgrepo-com.svg" width="20px" alt="statistics">
+              <img src="src/images/smiley.svg" width="20px" alt="smiley">
+              <img src="src/images/schedule.svg" width="20px" alt="schedule">
+              <button id="post-tweet">Tweet</button>
+            </div>
+        </div>
+
+        <div class="tweets-container">
```
extra-jalapenos, line 69: Sorry that I'm only now reviewing – had technical difficulties. Have you tested the scroll behaviour with a lot of tweets, that the sidebar and nav stay static?

homepage.html
```diff
@@ -0,0 +1,144 @@
...
+    <meta http-equiv="X-UA-Compatible" content="ie=edge"/>
+    <link rel="stylesheet" href="style.css"/>
+    <title>Twitter</title>
+  </head>
+  <body>
+    <div class="main-container">
+      <header class="header">
+        <img src="src/images/x-logo.svg" width="32px" alt="x logo">
+        <nav>
+          <div class="header-button">
```
extra-jalapenos, line 15: have you considered making this a ul instead of a lot of divs?

extra-jalapenos (commented):
Really like the clickable / fillable create new tweet box & search!

homepage.html
Radio58: Ah thank you, that's something I forgot to do

PR #125, opened 2023-09-28
Satoki

Satokii:


index.html
```diff
@@ -0,0 +1,461 @@
...
+                            <img class="new-message" src="./svgs/right-menu-svgs/new-message-icon.svg" alt="new message icon">
+                            <img class="expand" src="./svgs/right-menu-svgs/expand-menu.svg" alt="">
+                        </div> 
+                    </a>
+                </div>
+            </div>
+        </main>
+    </div>
+</body>
+</html>
```
TomEastwood, line 465: The amount of detail youve gone into in your HTML is awesome, and really clear to read too

index.html
Chloe070196, line 1: Awesome layout - the proportions look great

style.css
```diff
@@ -0,0 +1,828 @@
...
+    color: #03ba18;
+}
+
+/* LIKE REACTIONS */
+.like-reaction {
+    text-decoration: none;
+}
+
+.like-reaction:hover .like {
+    filter: invert(20%) sepia(62%) saturate(7458%) hue-rotate(318deg) brightness(85%) contrast(100%);
```
TomEastwood, line 483: The use of filters is super powerful and its cool to see how many of the features youve used

index.html
Chloe070196, line 1: line 102 - dunno if this is already planned, but I'd suggest making the create post bit stick to the top of the screen when the user is scrolling through the feed.

Chloe070196: also, you've probably noticed this already, but your ''what's happening" bit is currently displaying over multiple line instead of one

style.css
```diff
@@ -362,27 +364,128 @@
...
+    padding: 4px;
+    border-radius: 30px;
+}
+
+/* REPLY REACTIONS */
+.reply-reactions {
+    text-decoration: none;
+}
+
+.reply-reactions:hover .reply {
```
Chloe070196, line 446: your :hover properties are on point, methinks! :)) looks very twitter-y (well, I guess now 'x-y"? that just doesn't sound good.)

style.css
```diff
@@ -294,16 +296,16 @@
...
-}
 
 .main-feed > div {
     display: grid;
     grid-template-columns: 50px 1fr;
     grid-auto-rows: auto;
     padding: 15px 15px;
     border-bottom: solid 1px var(--border-color);
+    border-left: solid  1px var(--border-color);
+    border-right: solid  1px var(--border-color);
```
Chloe070196, line 362: Quite curious about the var(--border-color) bit - first time I've stumbled upon that! Mind explaining what its job is? Cheers!

index.html
```diff
@@ -369,46 +409,46 @@
...
-                                    <a class="follow-link" href="#">Follow</a>
+                                    <button class="follow-link">Follow</button>
                                 </div>
-                            </div>
-                            <div class="follow-user">
+                            </a>
+                            <a class="follow-user" href="#">
                                 <img class="follow-user-photo" src="https://yt3.ggpht.com/a/AGF-l78-OMGsDVO-YjQWPkm-w84cKf2SO-qBldcrRA=s900-c-k-c0xffffffff-no-rj-mo" alt="user photo">
                                 <div class="follow-user-details">
                                     <div class="follow-user-name">Daffy Duck</div>
```
Chloe070196: the duck theme is strong with this one - love it!

index.html
EwbiDev, line 1: 🦆🦆🦆🦆🦆🦆🦆🦆🦆🦆🦆🦆🦆🦆🦆🦆🦆

style.css
EwbiDev, line 1: Design looking cleeean!

index.html
Satokii: Thank you, yes this is next on my list of an ever growing number of things to do 😂

style.css
```diff
@@ -0,0 +1,828 @@
...
+}
+
+/* SPECIFIC POST ADJUSTMENTS */
+
+.post-1 .tweet-photo {
+    border-radius: 20px;
+    object-fit: cover;
+    width: 400px;
+    height: 400px;
+}
```
EwbiDev, line 552: We need some hover effects on these like you've done with `sub-card` on your trending menus! 😇

style.css
```diff
@@ -0,0 +1,828 @@
...
+/* OTHER ICON ADJUSTMENTS */
+.direct-tweet-reaction > a > div > .reply {
+    width: 28px;
+}
+
+.direct-tweet-reaction > a> div > .views {
+    width: 27px;
+}
+
+/* SPECIFIC POST ADJUSTMENTS */
```
EwbiDev, line 545: A lot of these feel as though they can be made more generalised into a .post class overall. It'll get tricky having to specifically tweak every post if you want to add a new one.

style.css
Satokii: So these are just global variables I've created. If you see at the top of my CSS the ' :root ' part is where I've made a few variables (just colours for this example) that I know I will be using multiple times throughout my code. You can then just refer back to that variable you created using var() later in the css, which is helpful if you create a specific style, colour, etc that may be unique. Hope this helps!

style.css
EwbiDev: css file was easy to follow, was able to grasp what you were changing easily. Comments helped with that too!

style.css
```diff
@@ -0,0 +1,828 @@
...
+    color: rgb(219, 219, 219);
+    max-height: 100vh;
+}
+
+
+:root {
+    --border-color: rgba(128, 128, 128, 0.632);
+    --box-background-color: rgb(32, 35, 39);
+    --hover-bg-color: rgba(106, 106, 106, 0.362);
+}
```
EwbiDev, line 17: Gotta love variables, nice one!

EwbiDev (commented):
Nice job! Really solid overall. Last comment is that you are often repeating the same properties over and over which you could try to solve by making some more general classes. For example `display: grid` appear a huge amount, may be worth making a `.grid` selector which has some base attributes which you can overwrite if necessary?

style.css
Satokii: Haha i'll add it to my to-do-list!

style.css
Satokii: Cheers, yeah good shout! I'll have a look into implementing this.

PR #129, opened 2023-09-29
Nazar Tymiv

NazarTymiv:
**Don't finish yet**


css/style.css
```diff
@@ -0,0 +1,292 @@
...
+    z-index: 2;
+}
+
+.header__logo::before {
+    content: '';
+    width: 180%;
+    height: 180%;
+    position: absolute;
+    border-radius: 100%;
+    transition: all 0.1s ease-in;
```
yee0802, line 63: Nice use of transition property. Makes your page look that extra bit better!

NazarTymiv:
Layout has finished just I should to do responsive layou and project will be finished :)

NazarTymiv:
Project has finished!

PR #131, opened 2023-09-29
Tom Eastwood

TomEastwood:


twitter.html
```diff
@@ -0,0 +1,220 @@
...
+                    </div>
+                </div>
+                <div class="post">
+                    <div class="feed">
+                        <img class="pfp" src="./images/pfp-1.jpg" alt="profile" />
+                        <p><b>Tom</b> @Tom</p>    
+                        <p class ="lower">check out this cool cat.</p>
+                    </div>
+                    <div class="pic">
+                        <img class="img" src="./images/cat.jpg" alt="cat" />
```
EwbiDev, line 137: This is indeed a cool cat.

twitter.css
```diff
@@ -0,0 +1,362 @@
...
+    height: 30px;
+    margin: 10px;
+}
+
+.upload-icons .logo:hover {
+    padding: 5px;
+    border-radius: 50%;
+    background-color: rgb(207, 207, 207);
+    transition: 0.5s;
+}
```
EwbiDev, line 189: These effects are joooocy 😍

twitter.css
```diff
@@ -0,0 +1,362 @@
...
+    grid-template-columns: 70px 1fr;
+    column-gap: 20px;
+    align-items: center;
+    font-size: 30px;
+    color: gray;
+    padding: 0 20px;
+}
+
+.upload-icons {
+    display: inline-flex;
```
EwbiDev, line 172: Boo, flex! *hssss*

twitter.css
EwbiDev, line 1: Selectors looking good, nothing looks hyper pointed at. All clear what they're meant to be doing! 🙂

twitter.css
```diff
@@ -0,0 +1,362 @@
...
+
+@media screen and (max-width: 1900px) {
+    .right-nav {
+        width: 200px;} 
+}
+
+@media screen and (max-width: 1400px) {
+    .right-nav {
+        display: none;}
+}
```
EwbiDev, line 275: This query seems to set off quite late, stuff is already getting cut off before this point.

twitter.html
```diff
@@ -0,0 +1,220 @@
...
+                            <img class="logo" src="./images/share.svg" alt="share" />
+                        </div>
+                    </div>
+                </div>
+            </div>
+        </div>
+        <div class="right-nav">
+            <div class="search">
+                <img class="logo" src="./images/search.svg" alt="search" />
+                <h2>Search Twitter</h2>
```
EwbiDev, line 159: I can see you have an `input` pointer in your css, you should use it somewhere!

twitter.html
```diff
@@ -0,0 +1,220 @@
...
+        </div>
+
+        <div class="main-body">
+            <div class="main-body-header">
+                <div class="title">
+                    <h2>Home</h2>
+                </div>
+                <div class="tweet-box">
+                    <img class="pfp" src="./images/pfp-1.jpg" alt="profile" />
+                    <p>What's happening?</p>
```
EwbiDev, line 64: As per search bar.

EwbiDev (commented):
Nice!

twitter.css
```diff
@@ -0,0 +1,362 @@
...
+    grid-template-rows: 100px 500px 500px;
+    grid-template-columns: 400px;
+    gap: 20px;
+    position: sticky;
+    top: 0;
+    padding: 20px;
+    background-color: white;
+}
+
+@media screen and (max-width: 1900px) {
```
NazarTymiv, line 267: Try to use all of the @media in the end of the css file it will be helpfull for you to reduce a lot of errors because sometimes @media can stop working if you are using them on the middle of the css file and mixing them with normal styles :)

twitter.html
```diff
@@ -0,0 +1,220 @@
...
+                    </div>
+                    <div class="icon">
+                        <img class="logo" src="./images/poll.svg" alt="poll" />
+                    </div>
+                    <div class="icon">
+                        <img class="logo" src="./images/emoji.svg" alt="emoji" />
+                    </div>
+                    <div class="icon">
+                        <img class="logo" src="./images/schedule.svg" alt="schedule" />
+                    </div>
```
NazarTymiv, line 81: I like this section because you are using container for all of the tags img :)

twitter.html
```diff
@@ -0,0 +1,220 @@
...
+                </div>
+                <div class="item">
+                    <img class="logo" src="./images/lists.svg"><a href="#">Lists</a>
+                </div>
+                <div class="item">
+                    <img class="logo" src="./images/profile.svg"><a href="#">Profile</a>
+                </div>
+                <div class="item">
+                    <img class="logo" src="./images/more.svg"><a href="#">More</a>
+                </div>
```
NazarTymiv, line 41: Here is some wrong because you are using tag a only for text and if you try to click on it it will be work only when you will click on the text and if you click on the icon which you put here it will not work. So try to create a tag and inside this tag put img and span tag for text.

```html
<a href="#">
    <img class="logo" src="./images/bookmark.svg">
    <span>Bookmarks<span>
</a>
```

NazarTymiv (commented):
Good job :)

twitter.css
TomEastwood: I literally tried every grid and display method and none of them worked but that one single one worked perfectly as in intended wher-as `display: inline-block;` didnt.

PR #134, opened 2023-09-29
Akindele Ayorinde

Hayor4real:


style.css
```diff
@@ -0,0 +1,308 @@
...
+
+.container {
+  display: grid;
+  grid-template:
+    ".   left__menu   main__content     right__menu   .  " 1000px /
+    100px 250px 1fr 250px 100px;
+  gap: 5px;
+}
+
+/* LEFT MENU*/
```
Chloe070196, line 13: I like the layout of your left-menu - it's neat and displays very nicely

index.html
```diff
@@ -0,0 +1,611 @@
...
+                  <path
+                    d="M7.646 1.146a.5.5 0 0 1 .708 0l3 3a.5.5 0 0 1-.708.708L8.5 2.707V11.5a.5.5 0 0 1-1 0V2.707L5.354 4.854a.5.5 0 1 1-.708-.708l3-3z"
+                  />
+                </svg>
+              </a>
+            </div>
+          </div>
+        </div>
+      </div>
+      <div class="right__menu">
```
Chloe070196, line 539: on my browser, this is displayed on the very left of the screen (in what I guess should be white space), instead of to the right of your main content.

style.css
```diff
@@ -0,0 +1,308 @@
+body {
+  font-family: Arial, Helvetica, sans-serif;
+}
+
+.container {
+  display: grid;
+  grid-template:
+    ".   left__menu   main__content     right__menu   .  " 1000px /
```
Chloe070196, line 8: the right__menu grid area is defined, but it looks like no element has been assigned to it - I'm guessing this is probably why your right menu is being display the empty container to the very left of the screen.

style.css
Chloe070196, line 1: not certain why, but reducing the width of the screen causes main to disappear, and leaves your left-menu at the centre of the screen, instead of having both left menu and right menu disappear, with main remaining. I reckon it might be worth having a look at making a new media query to help with that

style.css
```diff
@@ -0,0 +1,308 @@
...
+}
+
+@media screen and (min-width: 80rem) {
+  .main__content {
+    margin: 1em 2em;
+  }
+}
+
+/* MAIN CONTAINER */
+.main__content {
```
Chloe070196, line 69: Nice to have the feed as a scrollable element!

style.css
```diff
@@ -0,0 +1,308 @@
...
+  width: 70px;
+  height: 15px;
+  margin-top: 3px;
+  border-bottom: 2px solid lightgray;
+}
+
+.main__content .main__tweet {
+  display: grid;
+  border-top: lightgray solid 1px;
+  grid-template-rows: repeat(auto-fit, 1fr);
```
Chloe070196, line 159: Cool! I'm guessing that's to allow for any number of posts to be added to the feed? Nice feature!

style.css
Hayor4real: exactly i will fix still currently working on it thanks

PR #152, opened 2024-03-14
Periklis K

PerikK:


index.html
```diff
@@ -0,0 +1,1378 @@
...
+                </button>
+            </div>
+
+            <div class="spacer"></div>
+            <div class="spacer nav-link"></div>
+            <!-- <div class="spacer nav-link"></div> -->
+
+            <div class="user">
+                <div class="profile-pic">
+                    <img src="./svgFiles/Hello Ctulhu! black letters.jpg" alt="" width="45px">
```
vherus, line 150: I would avoid having spaces and special characters (!) in file names, they can cause unexpected issues on certain systems. Better to go with underscores and dashes just to be safe

index.html
```diff
@@ -0,0 +1,1378 @@
...
+                            <svg width="20" height="20" viewBox="0 0 24 24" aria-hidden="true" class="add-gif"
+                                style="color: rgb(29, 155, 240);">
+                                <g>
+                                    <path
+                                        d="M3 5.5C3 4.119 4.12 3 5.5 3h13C19.88 3 21 4.119 21 5.5v13c0 1.381-1.12 2.5-2.5 2.5h-13C4.12 21 3 19.881 3 18.5v-13zM5.5 5c-.28 0-.5.224-.5.5v13c0 .276.22.5.5.5h13c.28 0 .5-.224.5-.5v-13c0-.276-.22-.5-.5-.5h-13zM18 10.711V9.25h-3.74v5.5h1.44v-1.719h1.7V11.57h-1.7v-.859H18zM11.79 9.25h1.44v5.5h-1.44v-5.5zm-3.07 1.375c.34 0 .77.172 1.02.43l1.03-.86c-.51-.601-1.28-.945-2.05-.945C7.19 9.25 6 10.453 6 12s1.19 2.75 2.72 2.75c.85 0 1.54-.344 2.05-.945v-2.149H8.38v1.032H9.4v.515c-.17.086-.42.172-.68.172-.76 0-1.36-.602-1.36-1.375 0-.688.6-1.375 1.36-1.375z">
+                                    </path>
+                                </g>
+                            </svg>
+
+                            <!-- <div class="icon-spacer"></div> -->
```
vherus, line 224: Try not to commit dead code like this. We can always go back to previous commits, that's what git is for, so don't worry about keeping things you might need in future

index.html
```diff
@@ -0,0 +1,1378 @@
...
+                    <div class="name"></div>
+                    <div class="follow"></div>
+                </div> -->
+
+            </div>
+            <div class="trends">
+                <p class="trends-title">Trends</p>
+
+
+
```
vherus, line 1266: There's a lot of chaotic white space in this file...

index.html
```diff
@@ -0,0 +1,1378 @@
...
+    </div>
+
+
+
+
+
+
+
```
vherus, line 1375: oh

style.css
```diff
@@ -0,0 +1,731 @@
+* {
+    margin: 0;
+}
+
+.container {
+    /* height: 100vh; */
+    display: grid;
+    grid-template:
+        'header main news' /
+        25% 1fr 30%
```
vherus, line 10: It can be difficult to work with %'s for layouts, it's better to define fixed pixel widths for side bars and change them with media queries at certain break points

PerikK: Oh god! And I thought that %'s will be better for responsive design since analogy stays constant no matter what the actual size is. Duly noted as are the rest of your remarks.

PR #156, opened 2024-03-15
Sami Saeed

samisalehsaeed:


new_twitter.css
```diff
@@ -0,0 +1,243 @@
+*{
+    margin: 0;
+}
+.container{
+    display:grid;
+    height: 100vh;
+    grid-template:
+    'left-sidebar main right-sidebar'/300px 1fr 300px
+}
+.logo{
```
vherus, line 10: You should add an empty line after every CSS block to give some breathing space, it makes it a little easier to read / maintain

new_twitter.css
```diff
@@ -0,0 +1,243 @@
...
+    img{
+    height:40px;
+    width:47px;
+    padding-left:5px
+    }
```
vherus, line 15: Why is this `img` element inside of the `logo` element? You can't nest css selectors inside of others, except with things like media queries. You should use specificity to target specific child elements:

```css
.logo img {

}
```

new_twitter.css
```diff
@@ -0,0 +1,243 @@
...
+    padding-left: 15px;
+    border: none;
+    cursor: pointer;
+    background-color: white;
+    font-size: 15px;
+    font-family: 'Gill Sans', 'Gill Sans MT', Calibri, 'Trebuchet MS', sans-serif;
+    font-weight: 700;
+    img{
+        height: 25px;
+    }
```
vherus, line 43: Same here, you can't nest this img element inside the home-btn element

new_twitter.html
```diff
@@ -0,0 +1,72 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="stylesheet" href="./new_twitter.css" />
+</head>
+<body>
+    <div class = 'container'>
```
vherus, line 9: Don't add white space around the property/value pairs. It works, but it goes against convention - I don't know any project that uses whitespace in this way. Also, be sure to use double quotes instead of single quotes

`<div class="container">`

vherus (commented):
Good work! Be careful with nesting CSS selectors inside of other CSS selectors

PR #158, opened 2024-03-15
Farshad Bagdeli

FBagdeli:
I still need to work on it


home-style.css
```diff
@@ -0,0 +1,467 @@
...
+  background-color: #000000;
+
+  svg {
+    padding: 10px;
+    height: 40px;
+    width: 40px;
+    fill: white;
+  }
+
+}
```
vherus, line 36: You can't nest CSS selectors inside of other selectors, you'd need to use specificity:

```css
.logo {
  grid-area: logo;
  background-color: #000000;
}

.logo svg {
    padding: 10px;
    height: 40px;
    width: 40px;
    fill: white;
}
```

home-style.css
```diff
@@ -0,0 +1,467 @@
...
+    }
+    svg {
+      fill: white;
+      height: 30px;
+      width: 30px;
+      padding: 15px;
+    }
+  }
+  
+}
```
vherus, line 63: Same here, and for a lot of your CSS actually. I imagine this has been causing you a bit of a headache.

vherus (commented):
You HTML is good, and your CSS is fine but you're nesting a lot of things in CSS. CSS shouldn't have any nesting other than with things like media queries

FBagdeli:
I will fix it and try to change Div's With right semantics also

Thanks in advance for your guidance
________________________________

FBagdeli:
@vherus 
Will work on it and try to change the html with right semantics also

Thanks in advance

PR #159, opened 2024-03-15
JOAN NABUUMA1Twitter-clone-challenge

JOANNABUUMA1:
Solutions for both the assessment and extended criteria.


html-twitter-challenge/style.css
```diff
@@ -0,0 +1,408 @@
...
+
+
+
+
+
+
+
+
+@media (max-width:1920) {
+    .desktop\:hidden {
```
vherus, line 323: I'm not sure why there's a `\` here, is that intentional?

vherus (commented):
Joan, this is really, really great work. Well done! You've definitely turned things around since the rocky start, congrats! Keep this up :)

PR #162, opened 2024-03-15
Shaun Harris

Shaun-Harris:
Not having the responsiveness I would like.


twitter.html
```diff
@@ -0,0 +1,203 @@
...
+                <div class="messages"><svg fill="#E7E9EA" viewBox="0 0 24 24" aria-hidden="true" class="logo-svg"><g><path d="M1.998 5.5c0-1.381 1.119-2.5 2.5-2.5h15c1.381 0 2.5 1.119 2.5 2.5v13c0 1.381-1.119 2.5-2.5 2.5h-15c-1.381 0-2.5-1.119-2.5-2.5v-13zm2.5-.5c-.276 0-.5.224-.5.5v2.764l8 3.638 8-3.636V5.5c0-.276-.224-.5-.5-.5h-15zm15.5 5.463l-8 3.636-8-3.638V18.5c0 .276.224.5.5.5h15c.276 0 .5-.224.5-.5v-8.037z"></path></g></svg><p>Messages</p></div>
+                <div class="lists"><svg fill="#E7E9EA" viewBox="0 0 24 24" aria-hidden="true" class="logo-svg"><g><path d="M3 4.5C3 3.12 4.12 2 5.5 2h13C19.88 2 21 3.12 21 4.5v15c0 1.38-1.12 2.5-2.5 2.5h-13C4.12 22 3 20.88 3 19.5v-15zM5.5 4c-.28 0-.5.22-.5.5v15c0 .28.22.5.5.5h13c.28 0 .5-.22.5-.5v-15c0-.28-.22-.5-.5-.5h-13zM16 10H8V8h8v2zm-8 2h8v2H8v-2z"></path></g></svg><p>Lists</p></div>
+                <div class="bookmarks"><svg fill="#E7E9EA" viewBox="0 0 24 24" aria-hidden="true" class="logo-svg"><g><path d="M4 4.5C4 3.12 5.119 2 6.5 2h11C18.881 2 20 3.12 20 4.5v18.44l-8-5.71-8 5.71V4.5zM6.5 4c-.276 0-.5.22-.5.5v14.56l6-4.29 6 4.29V4.5c0-.28-.224-.5-.5-.5h-11z"></path></g></svg><p>Bookmarks</p></div>
```
vherus, line 26: This is incredibly difficult to read and maintain, I can't really tell where one SVG begins and another ends. You should try to keep things formatted nicely so it's easy to read:

```html
<div class="logo">
    <svg fill="#E7E9EA" viewBox="0 0 24 24" aria-hidden="true" class="x-svg">
        <g>
            <path d="M18.244 2.25h3.308l-7.227 8.26 8.502 11.24H16.17l-5.214-6.817L4.99 21.75H1.68l7.73-8.835L1.254 2.25H8.08l4.713 6.231zm-1.161 17.52h1.833L7.084 4.126H5.117z"></path>
        </g>
    </svg>
</div>
```

twitter.html
```diff
@@ -0,0 +1,203 @@
...
+
+                
+
+            </div>
+        
+
+            <div class="show">Show 243 posts</div>
+
+        
+
```
vherus, line 77: I don't think we need this much white space

vherus (commented):
Your work is good but I think you could tidy it up a bit, this would be a little difficult to maintain in a real project

twitter.html
Shaun-Harris: Hi Nathan, Will take this on board. I think in this case all the <div>'s were confusing me in in terms of alignment (which I found easier if on one line). However, I think today's lesson should eliminate that by using different semantics. So I know and can clearly see that there is an opening and closing to the semantics by having varied names rather than consistent <div>s  which make it harder to read. 

Do you want me to alter this file and resubmit or just apply this feedback to work going forward?

twitter.html
Shaun-Harris: Again I think this is a tool for me as I have dyslexia and sometimes require the space. If you have a better solution that could be helpful such as text size etc. that would be great.

vherus: Ah, you could try going to File -> Preferences -> Settings in VS Code and finding the Editor: Font Size setting. Bump that up to a value you feel comfortable with, hopefully that may help things.

You could also continue to use this white space approach while developing the solution, and then just remove it all when you're done so you still get the helpfulness of it and it doesn't make its way into the final product

twitter.html
vherus: No need to update this, just feedback to keep in mind going forward :)

PR #165, opened 2024-03-18
Mamadou Leye

homonoviscoding:


index.html
```diff
@@ -0,0 +1,179 @@
...
+    <!-- feed starts -->
+    <div class="feed">
+      <div class="feed__header">
+        <h2>Home</h2>
+      </div>
+
+      <!-- tweetbox starts -->
+      <div class="tweetBox">
+        <form>
+          <div class="tweetbox__input">
```
vherus, line 73: Be careful with case sensitivity. This class is `tweetbox`, but the others are `tweetBox`.

Stick with kebab-case for CSS, not camelCase

styles.css
```diff
@@ -0,0 +1,243 @@
+* {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
+
+body {
+  --twitter-color: #50b7f5;
+  --twitter-background: #e6ecf0;
```
vherus, line 9: Nice use of variables

vherus (commented):
Great work! Your code is easy to read, well constructed and mostly consistent.

homonoviscoding:
Thank you, will use kebab case from now on. I just wanted to experiment
things I have never done before with the challenge.

PR #168, opened 2024-03-18
Oluwagbemi Abiodun

Alakowe19:


index.html
```diff
@@ -0,0 +1,176 @@
...
+</head>
+<body>
+    <!-- building side, it starts here -->
+    <div class="sidebar">
+        <i class="fab fa-twitter"></i>
+
+    <div class="sidebarOption active">
+        <span class="material-icons"> home </span>
+        <h2>Home</h2>
+        </div>
```
vherus, line 23: Your indentation is a little off, I'm not sure where one div ends and another begins. This would be easier to maintain like this:


```html
    <div class="sidebar">
        <i class="fab fa-twitter"></i>

        <div class="sidebarOption active">
            <span class="material-icons"> home </span>
            <h2>Home</h2>
        </div>
```

index.html
```diff
@@ -0,0 +1,176 @@
...
+            </div>
+        </div>
+    </div>
+
+   </div>
+<!-- Post sector ends here -->
+
+
+<!-- I start buiding my post section here -->
+<div class="post">
```
vherus, line 121: I don't think this div should be floating all the way at the left side of the file

vherus (commented):
Great work. Your indentation needs a bit of work to improve its consistency and quality, and your css should be using kebab-case instead of camelCase. Keep up the great work :)

